PR #2, opened 2017-04-20
Cache

delvedor:
Now `lookup` caches all the routes and there is a sensible improvement of performances :)

@mcollina feedbacks?

### DIFF
--- a/index.js
+++ b/index.js
@@ -138,15 +138,16 @@ Router.prototype.lookup = function (req, res) {
   var i = 0
   var len = req.url.length
   while (i < len && req.url[i] !== '?' && req.url[i] !== '#') i++
+  var url = req.url.slice(0, i)
 
   // search the route inside the cache
-  var handle = this.cache.get(req.url.slice(0, i))
+  var handle = this.cache.get(url)
   if (handle) return handle.handler(req, res, handle.params, handle.store)
 
   // if the route in not saved in the cache, we search it
-  handle = this.find(req.method, req.url.slice(0, i))
+  handle = this.find(req.method, url)
   if (!handle) return this._defaultRoute(req, res)
-  this.cache.set(req.url.slice(0, i), handle)
+  this.cache.set(url, handle)
   return handle.handler(req, res, handle.params, handle.store)
 }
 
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -132,11 +134,19 @@ Router.prototype._insert = function (method, path, kind, params, handler, store)
...
-  var handle = this.find(req.method, req.url.slice(0, i))
+
+  // search the route inside the cache
+  var handle = this.cache.get(req.url.slice(0, i))
+  if (handle) return handle.handler(req, res, handle.params, handle.store)
+
+  // if the route in not saved in the cache, we search it
+  handle = this.find(req.method, req.url.slice(0, i))
   if (!handle) return this._defaultRoute(req, res)
+  this.cache.set(req.url.slice(0, i), handle)
```
mcollina: you can set the result of `req.url.slice(0, i)` in a variable, you are computing the same thing everywhere.

delvedor: I did it, but the benchmark didn't change, this because we are computing the `slice` only if we have a *cache miss*, so one time per route.

mcollina: it's still a code smell. It's easy to make a mistake when updating this code.

mcollina: In a real-world scenario, your cache will likely be crunching for dynamic routes.
You should verify the performance with a dynamic parameter that is changing all the time, causing a cache miss _for every request_.

delvedor: You are right, if I run this benchmark
```js
.add('lookup dynamic route - cache random path', function () {
  findMyWay.lookup({ method: 'GET', url: '/user/' + i++ }, null)
})
```
I got a 8x-10x perf loss.

We can make the cache optional and warn the user about this behaviour. So if the user will have a lot of static routes can use the caching system, otherwise he can disable it.

mcollina: I think the added complexity would remove any benefit we could get from the caching.

delvedor: After a lot of benchmarks I must admit you are right, the optional caching helps a lot, but at the end in a real world context (fastify, for example), this change does not help in a sensible way.
Maybe we can figure out a way to make the lookup/find perform better.

PR #3, opened 2017-04-21
Automatically decode the uri

delvedor:
As titled.

In addition has been added a test case for the update of the store, the docs has been updated as well.

cc @mcollina


mcollina:
Does this affecta the benchmarks anyhow?

delvedor:
Yes, around 6%-7% less operation per second.

mcollina:
LGTM.

PR #6, opened 2017-05-08
replace string equality with charCode equality

billouboq:
Testing benchmark on node 7.10 : 

Before : 

lookup static route x 14,919,746 ops/sec ±0.77% (90 runs sampled)
lookup dynamic route x 662,481 ops/sec ±0.63% (94 runs sampled)
lookup long static route x 2,311,837 ops/sec ±0.98% (90 runs sampled)
find static route x 28,791,481 ops/sec ±0.88% (87 runs sampled)
find dynamic route x 732,577 ops/sec ±0.64% (94 runs sampled)
find long static route x 6,701,436 ops/sec ±0.75% (91 runs sampled)

After :

lookup static route x 16,330,523 ops/sec ±0.81% (91 runs sampled)
lookup dynamic route x 980,184 ops/sec ±0.61% (94 runs sampled)
lookup long static route x 3,957,358 ops/sec ±0.70% (90 runs sampled)
find static route x 29,006,918 ops/sec ±0.83% (94 runs sampled)
find dynamic route x 1,047,862 ops/sec ±0.53% (92 runs sampled)
find long static route x 6,818,069 ops/sec ±0.73% (91 runs sampled)

### DIFF
--- a/index.js
+++ b/index.js
@@ -37,12 +37,12 @@ Router.prototype.on = function (method, path, handler, store) {
   for (var i = 0, len = path.length; i < len; i++) {
     // search for parametric or wildcard routes
     // parametric route
-    if (path[i] === ':') {
+    if (path.charCodeAt(i) === 58 /* : */) {
       j = i + 1
       this._insert(method, path.slice(0, i), 0, null, null, null)
 
       // isolate the parameter name
-      while (i < len && path[i] !== '/') i++
+      while (i < len && path.charCodeAt(i) !== 47 /* / */) i++
       params.push(path.slice(j, i))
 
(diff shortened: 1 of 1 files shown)


billouboq:
@delvedor I see no improvement with this particular change.

delvedor:
That's strange!
Since it helps here
```js
 while (i < req.url.length && req.url.charCodeAt(i) !== 63 /* ? */ && req.url.charCodeAt(i) !== 35 /* # */) i++
```
I was expecting the same result. Thanks for having tried it!

delvedor:
Released in `v0.2.3` 🎉

PR #7, opened 2017-05-20
optimize lookup

billouboq:
avoid unecessary slice of url and I think it's also easier to read.

Using node JS 7.10 on mac OS :

From

lookup static route x 16,199,148 ops/sec ±0.61% (92 runs sampled)
lookup dynamic route x 945,503 ops/sec ±0.54% (93 runs sampled)
lookup long static route x 4,024,544 ops/sec ±0.47% (93 runs sampled)

To

lookup static route x 21,672,160 ops/sec ±0.72% (91 runs sampled)
lookup dynamic route x 1,039,106 ops/sec ±0.92% (92 runs sampled)
lookup long static route x 4,241,299 ops/sec ±0.59% (92 runs sampled)

### DIFF
--- a/index.js
+++ b/index.js
@@ -132,9 +132,7 @@ Router.prototype._insert = function (method, path, kind, params, handler, store)
 }
 
 Router.prototype.lookup = function (req, res) {
-  var i = 0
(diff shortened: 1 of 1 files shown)


mcollina:
🎉🎉🎉🎉

PR #9, opened 2017-06-15
Browser support

delvedor:
As titled! 🎉


mcollina:
I think you should run the browser test in a browser, check out tape and tape-run.

Why the completely different codebase?

delvedor:
> I think you should run the browser test in a browser, check out tape and tape-run.

👍 

> Why the completely different codebase?

Because we are not using the method and the lookup is different, also the `Node` class is slightly different. I didn't want to add overhead trying to understand if we are in the browser environment or not.

delvedor:
I don't want to drop `tap` as test runner, I'll create a new repo with the browser implementation of the router. :)

alexkrolick:
Is the use case for the browser router as a primitive for creating client-side routers? Eg: Listen to history navigation events and call the `router.lookup()` with the path?

Issue #15, opened 2017-07-07
Using regex to capture the URL parameters


gajus:
The following route definition works in Express.js, but does not work in Fastify/ find-my-way
, `/cinema-logos/:cinemaId(\\d+)-:cinemaName.svg`.

In Fastify, it produces `params: { 'cinemaId(\d+)-:cinemaName.svg': '1000146-vue.svg' },`, whereas I expect to get `cinemaId` and `cinemaName` as separate parameters.

`:cinemaId-:cinemaName` is not working either, because the matching expression appears to be greedy: `'cinemaId-:cinemaName.svg': '1000146-vue.svg'`.

gajus:
Looks like only one parameter can be used per "directory", https://github.com/delvedor/find-my-way/blob/41bdc50f823c47a6c716b746c7cf9dd0bb723ca7/index.js#L53-L54

This is very limiting ..

mcollina:
Thanks for reporting this! We should definitely have a look on how to best handle support for multiple dynamic parts of the URL.

Do you see a quick fix for this?

gajus:
Not unless you allow to use a regex in the URIs.

mcollina:
@gajus I do not think that will work, given how this library works.
However, `:cinemaId-:cinemaName` might do?

gajus:
You could just allow regex as the route definition. That would be an acceptable option, and it would be straightforward to implement, e.g.

```js
app.get('/cinema-logos/:cinemaId(\\d+)-:cinemaName([a-z0-9-]+).svg', ...

```

would become

```js
app.get(/\/cinema-logos\/(?<cinemaId>\d+)-(?<cinemaName>[a-z0-9-]+).svg/, ...

```

Named capture groups are stage 3, https://github.com/tc39/proposal-regexp-named-groups.

gajus:
> @gajus I do not think that will work, given how this library works.
> However, :cinemaId-:cinemaName might do?

You'd still need to allow to define regex for the parameter matching groups. That wouldn't be neither more performant or more convenient.

mcollina:
Given how it is implemented, I do not think this library could easily support regexps. Feel free to try a PR if you think it's easy or feasible.

gajus:
I've gone back to using Express.js as this issue is a deal-breaker and I doubt you are going to have fix soon. Raised an issue for reference.

mcollina:
Thanks for reporting!

delvedor:
[...]

PR #16, opened 2017-07-07
Basic regex support

delvedor:
With this PR we introduce a basic support for regex in urls (#15).
Example:
```js
router.on('GET', '/test/:id(^\\d+$)', (req, res, params) => {
  // your logic
})
```
Currently the following is not supported:
`/test/:id(^\\d+$).png` and `/test/:id(^\\d+$)-:param(^\\d+$)`.

In Express [path-to-regexp](https://github.com/pillarjs/path-to-regexp) is used, but is not working fine for us since we are using a completely different data structure for performances reasons (same reason we are avoiding regex in the first place).

Feedbacks?
If you have some suggestion to support also the other cases, please let me know.

*ps at the moment this pr is not introducing any performances regression.*

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,12 @@ router.on('GET', '/example/:name', () => {}))
 router.on('GET', '/other-example/*', () => {}))
 ```
 
+Regex routes are supported as well, but pay attention, regex are very expensive!
+```js
+// parametric with regex
+router.on('GET', '/test/:file(^\\d+).png', () => {}))
+```
+
 <a name="shorthand-methods"></a>
 ##### Shorthand methods
 If you want an even nicer api, you can also use the shorthand methods to declare your routes.
--- a/index.js
+++ b/index.js
@@ -5,6 +5,7 @@
     static: 0,
     param: 1,
     matchAll: 2,
+    regex: 3
 
   Char codes:
     '/': 47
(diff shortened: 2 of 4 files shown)


mcollina:
I think supporting /test/:id(^\\d+$).png  is feasible by editing the previous regexp by adding something after the matching group. Supporting the second might be a bit harder, and we might do that in a separate PR.

Basically we must extract two variables from the same part of the URL. Is that feasible, or it is a major shift in the internals?

delvedor:
> I think supporting /test/:id(^\d+$).png is feasible by editing the previous regexp by adding something after the matching group. Supporting the second might be a bit harder, and we might do that in a separate PR.

👍 

> Basically we must extract two variables from the same part of the URL. Is that feasible, or it is a major shift in the internals?

It will surely require some work, an idea could be split on `'-'` and add two parameters and regexes to the node and handle it at runtime, but it will be very slow.

mcollina:
I do not think this is a huge problem, as long as we can fix the .png thing. If we can match based on a specific regex, further parts of that can then be extracted by hand/using individual regexes.

delvedor:
Actually the *.png* thing could be solved with a slightly complex regex, for example:
```js
/[^\\]*\.(\w+)$/.test('image.png') // true
/[^\\]*\.(\w+)$/.test('image.jpg') // true

/[^\\]*\.(png)$/.test('image.png') // true
/[^\\]*\.(png)$/.test('image.jpg') // false
```

mcollina:
I meant that we should _edit_ the `/test/:id(^\\d+$).png` regexp, and add the trailing part.

delvedor:
It is already happening, because the following tests are passing:
```js
test('route with an extension regex', t => {
  t.plan(1)
  const findMyWay = FindMyWay({
    defaultRoute: () => {
      t.fail('route not matched')
    }
  })

  findMyWay.on('GET', '/test/:file(^\\d+).png', () => {
    t.ok('regex match')
  })

  findMyWay.lookup({ method: 'GET', url: '/test/12.png' }, null)
})

test('route with an extension regex - no match', t => {
  t.plan(1)
  const findMyWay = FindMyWay({
    defaultRoute: () => {
      t.ok('route not matched')
    }
  })

  findMyWay.on('GET', '/test/:file(^\\d+).png', () => {
    t.fail('regex match')
  })

  findMyWay.lookup({ method: 'GET', url: '/test/aa.png' }, null)
})
```
*All add this in the next commit*

Issue #20, opened 2017-08-07
Parametric path resolution


deividasjackus:
```js
const http = require("http");
const router = require("find-my-way")();

router.on("GET", "/a/:param", (req, res, params) => res.end(JSON.stringify(params)));

http.createServer(router.lookup.bind(router)).listen(3000);
```

### Works as expected
http://localhost:3000/a/perfectly-fine-route => `{"param":"perfectly-fine-route"}`

### Does not work
- http://localhost:3000/a => `{"param":"a"}`
- http://localhost:3000/a-non-existing-route => `{"param":"a-non-existing-route"}`
- http://localhost:3000/a/ => `TypeError: Cannot read property 'length' of null`

delvedor:
I'm able to reproduce this, thank you for reporting!
I'll fix it as soon as I can.

If you want to try, fire a PR! :)

mlovrovich:
```js
    const http = require('http')
    const router = require('find-my-way')()

    router.on('GET', '/:namespace/:type/:id', (req, res, params) => {
        res.end('{"message":"hello world"}')
    })

    router.on('POST', '/:namespace/jobs/:name/run', (req, res, params) => {
        res.end('{"message":"hello world"}')
    })

    const server = http.createServer((req, res) => {
        router.lookup(req, res)
    })

    server.listen(4000, err => {
        if (err) throw err
        console.log('Server listening on: http://localhost:4000')
    })
```

### Works as expected
GET http://localhost:4000/test_namespace/test_type/test_id
GET http://localhost:4000/test_namespace/jobss/test_id

### Does not work
GET http://localhost:4000/test_namespace/jobs/test_id

PR #29, opened 2017-09-27
Fix 28

delvedor:
As titled :)

@BigMurry can you also check if it fixes your issue?

This pr also changes the behavior of https://github.com/delvedor/find-my-way/issues/20 (and https://github.com/delvedor/find-my-way/pull/21), in the case of `/test/:param` with the route `/test/`.

There are not sensible changes in the performances 🎉

### DIFF
--- a/index.js
+++ b/index.js
@@ -203,19 +203,19 @@ Router.prototype.find = function (method, path) {
     // found the route
     if (pathLen === 0 || path === prefix) {
       var handle = currentNode.getHandler(method)
-      if (!handle) return null
+      if (handle) {
+        var paramNames = handle.params
+        var paramsObj = {}
 
-      var paramNames = handle.params
-      var paramsObj = {}
+        for (i = 0; i < paramNames.length; i++) {
+          paramsObj[paramNames[i]] = params[i]
+        }
 
-      for (i = 0; i < paramNames.length; i++) {
-        paramsObj[paramNames[i]] = params[i]
-      }
-
-      return {
-        handler: handle.handler,
-        params: paramsObj,
-        store: handle.store
+        return {
+          handler: handle.handler,
+          params: paramsObj,
+          store: handle.store
+        }
       }
     }
 
@@ -228,7 +228,7 @@ Router.prototype.find = function (method, path) {
       pathLen = path.length
     }
 
-    node = currentNode.find(path[0])
+    node = currentNode.find(path[0], method)
     if (!node) return null
     kind = node.kind
 
--- a/node.js
+++ b/node.js
@@ -42,10 +42,21 @@ Node.prototype.findByLabel = function (label) {
   return null
 }
 
-Node.prototype.find = function (label) {
+// Check in two different places the numberOfChildren and the map object
+// gives us around ~5% more speed
+Node.prototype.find = function (label, method) {
   for (var i = 0; i < this.numberOfChildren; i++) {
     var child = this.children[i]
-    if (child.map || child.children.length) {
+    if (child.numberOfChildren !== 0) {
+      if (child.label === label && child.kind === 0) {
+        return child
+      }
+      if (child.kind > 0) {
+        return child
+      }
+    }
+
+    if (child.map && child.map[method]) {
       if (child.label === label && child.kind === 0) {
         return child
       }
--- a/test/issue-20.test.js
+++ b/test/issue-20.test.js
@@ -61,20 +61,20 @@ test('Should be 404 / 3', t => {
     t.fail('We should not be here')
   })
 
-  findMyWay.lookup({ method: 'GET', url: '/a/' }, null)
+  findMyWay.lookup({ method: 'GET', url: '/a//' }, null)
 })
 
-test('Should be 404 / 4', t => {
+test('Should get an empty parameter', t => {
   t.plan(1)
   const findMyWay = FindMyWay({
     defaultRoute: (req, res) => {
-      t.pass('Everything good')
+      t.fail('We should not be here')
     }
   })
 
   findMyWay.on('GET', '/a/:param', (req, res, params) => {
-    t.fail('We should not be here')
+    t.equal(params.param, '')
   })
 
-  findMyWay.lookup({ method: 'GET', url: '/a//' }, null)
+  findMyWay.lookup({ method: 'GET', url: '/a/' }, null)
 })
--- a/test/issue-28.test.js
+++ b/test/issue-28.test.js
@@ -1,0 +1,118 @@
+'use strict'
+
+const t = require('tap')
+const test = t.test
+const FindMyWay = require('../')
+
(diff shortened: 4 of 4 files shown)


delvedor:
@mcollina already done!
> There are not sensible changes in the performances 🎉

BigMurry:
The original problem is solved. But I found a new one related to wildcard route. Given the following declaration:
```
findMyWay.on('GET', '/test1/foo', () => {}); 
findMyWay.on('GET', '/test2/foo', () => {});
findMyWay.on('OPTIONS', '/*', () => {});
```
When I request to `OPTIONS /test1/foo`, I will get `404`.
Cause in the `tree`, we will have the following prefix tree structure: 
```
{
 '/': {
    'test': {
      '1/foo': {},
      '2/foo': {}
    }, // branch 1: numberOfChildren !== 0
    '*': {} // branch 2
  }
}
```
As a result, when we request to `OPTIONS /test1/foo`, it will be navigated to `branch 1` and it will never have the chance to enter the `branch 2`. So we will not find the `/*` path. This also explain why the test case `Wildcard inside a node with a static route but different method` is passing. Cause for that route declaration, we'll get the following prefix tree:
```
{
  '/': {
    'test/hello': {}, // this branch will not match the http method condition, so it'll hit the wildcard branch
    '*': {}
  }
}
```

brunoscopelliti:
@BigMurry Good spot.

I don't know well the project as @delvedor .
The only possible way to solve this kind of issue that comes to my mind, is by evaluating the HTTP method upfront.

What do you think about creating a diffent prefix tree for each http method? Something like (in the constructor, or even lazily)

```
this.tree = {
  'GET':  new Node(),
  'OPTIONS':  new Node()
}
```

Then in the lookup we can have something like:

```
Router.prototype.find = function (method, path) {
  var currentNode = this.tree[method]
```
I do not have the time to attempt a PR right now... but I think it should solve the issue, and not make performance worst. Thoughts?

delvedor:
@BigMurry nice catch!
@brunoscopelliti I was thinking the same thing, because add a check for the method of the children inside `find` could be too expensive (especially because we don't know how deep we should go).
Build a different radix tree for every method could solve the issue with a very small cost, I'll work on it :)

brunoscopelliti:
Hi @delvedor,
I sketched the proposed solution on a branch (not yet public).

I can confirm that it solves the issue; however it requires a rework of the `prettyPrint` method... and have still to figure out the best way to approach the problem.
I would be happy to continue the work on this.

I've also run the benchmack.

*Proposed solution*

```
lookup static route x 15,147,002 ops/sec ±14.11% (91 runs sampled)
lookup dynamic route x 1,133,700 ops/sec ±0.88% (92 runs sampled)
lookup dynamic multi-parametric route x 718,804 ops/sec ±0.91% (90 runs sampled)
lookup dynamic multi-parametric route with regex x 600,590 ops/sec ±0.63% (89 runs sampled)
lookup long static route x 3,370,021 ops/sec ±0.77% (95 runs sampled)
find static route x 24,629,176 ops/sec ±0.86% (92 runs sampled)
find dynamic route x 1,223,641 ops/sec ±0.56% (90 runs sampled)
find dynamic multi-parametric route x 760,393 ops/sec ±0.75% (90 runs sampled)
find dynamic multi-parametric route with regex x 614,044 ops/sec ±1.75% (93 runs sampled)
find long static route x 4,722,067 ops/sec ±0.69% (94 runs sampled)
```

*Master*

```
lookup static route x 18,045,349 ops/sec ±0.99% (90 runs sampled)
lookup dynamic route x 1,112,466 ops/sec ±0.65% (92 runs sampled)
lookup dynamic multi-parametric route x 705,904 ops/sec ±0.81% (89 runs sampled)
lookup dynamic multi-parametric route with regex x 603,538 ops/sec ±2.52% (89 runs sampled)
lookup long static route x 3,344,521 ops/sec ±0.67% (91 runs sampled)
find static route x 25,018,652 ops/sec ±0.90% (93 runs sampled)
find dynamic route x 1,179,849 ops/sec ±0.55% (93 runs sampled)
find dynamic multi-parametric route x 751,334 ops/sec ±0.96% (93 runs sampled)
find dynamic multi-parametric route with regex x 618,492 ops/sec ±0.51% (91 runs sampled)
find long static route x 4,719,660 ops/sec ±0.78% (93 runs sampled)
```

It seems that the extra check I added at the begging of the `find` method, it's a bit slower for short static route.

```js
Router.prototype.find = function (method, path) {
  var currentNode = this.tree[method]
  if (!currentNode) {
    return null
  }
  ...
```

delvedor:
Hi @brunoscopelliti!
I'm working as well on that, it fixes the issue, but we fail in the case below:
```js
router.on('GET', '/test1/foo', (req, res, params) => {
  console.log('GET 1')
  res.end(req.url)
})

router.on('GET', '/test2/foo', (req, res, params) => {
  console.log('GET 2')
  res.end(req.url)
})

router.on('GET', '/*', (req, res, params) => {
  console.log('GET WILDCARD')
  res.end(req.url)
})
```
Withe the url `/test/foo`.

I found a solution, I've refactored `find` using a recursive strategy, in this way we can move between nodes on the same level.
At the moment I'm working on optimizing that solution :)

brunoscopelliti:
I see... I didn't consider that case.
This is another test case that came to my mind after having read your:

```
└── /
    ├── foo
    │   ├── 1/bar
    │   │   ├── 1/baz (GET)
    │   │   └── 2/baz (GET)
    │   └── 2/bar2/baz (GET)
    └── * (GET)
"

findMyWay.lookup({ method: 'GET', url: '/foo1/bar1/kuux' }, null)
// * (GET)
```

>  I've refactored find using a recursive strategy, in this way we can move between nodes on the same level.

Does your solution cover this one as well?
In this case we should be able not only to moves between siblings, but also to move one (or more...) level up.

To solve also this issue I'm thinking of making `Node.prototype.find` returns not just the first matching node it reaches, but eventually wildcard nodes.
If you're ok, I'll try working on a different approach.

delvedor:
I wrote the following test to reproduce your example, tell me if I'm wrong :D
```js
test('Mixed wildcard and static with same method', t => {
  t.plan(1)
  const findMyWay = FindMyWay({
    defaultRoute: (req, res) => {
    }
  })

  findMyWay.on('GET', '/foo1/bar1/baz', (req, res, params) => {
  })

  findMyWay.on('GET', '/foo1/bar2/baz', (req, res, params) => {
  })

  findMyWay.on('GET', '/foo2/bar2/baz', (req, res, params) => {
  })

    t.is(req.url, '/foo1/bar1/kuux')
  })

  findMyWay.lookup(
    { method: 'GET', url: '/foo1/bar1/kuux' },
    null
  )
})
```
With the recursive strategy the test passes correctly.

brunoscopelliti:
@delvedor Yes, that was the test case. Glad it works now.

PR #30, opened 2017-09-29
 Added url validations

delvedor:
As titled.

An url should start with `/` or `*` and could not be empty.

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,7 +16,6 @@
     '/': 47
     ':': 58
     '?': 63
-
 */
 
 const assert = require('assert')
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -300,9 +304,6 @@ Router.prototype.find = function (method, path) {
       path = path.slice(i)
       continue
     }
-
-    // route not found
-    if (len !== prefixLen) return null
```
mcollina: are we sure about this change?

delvedor: Yes, that code was forgotten there, its position is [here](https://github.com/delvedor/find-my-way/blob/e765c5b0df43049e03bdaf3b989fce0399713ab9/index.js#L245) :D

PR #27, opened 2017-09-23
Hanlde multi-parametric route (issue #17)

brunoscopelliti:
Detect multi parametric path (new nodeType=4)

Handle lookup for multi parametric path

Unit test issue #17

Issue #17, opened 2017-07-13
support for multiple params


mcollina:
I think the current problem for #15 is not about regexps. It's about multiple params in the same part:

```
/hello/:my-:world.html
```

Currently, find-my-way only supports one parameter within two `/`.

delvedor:
Our first goal is to not compromise the performances, and a feature like this would do it.

The only solution I see now is to discover if we are adding a *multiple parametric route* during the insertion and create a new kind of node, in this way at runtime we will not spent time trying splitting the parameter on `'-'`, but we'll do that only if we are in a *multiple parametric* node. In this way the performance impact will be low, just another `if` check.

The problem is how design the node object, since we are working with a radix tree, having multiple parameter could be hard to store and handle.

brunoscopelliti:
Hi @delvedor ,
I've tried to add support for multi parametric routes. I've followed your suggestion, and created a new node type for this kind of route.
Any feedback is appreciated.

delvedor:
Closed in #27

PR #31, opened 2017-10-01
Fix regression issue #17

brunoscopelliti:
Fix regression reported in https://github.com/delvedor/find-my-way/pull/27#issuecomment-333331583.

@AVVS, I'm sorry for the leaked breaking change.

### DIFF
--- a/index.js
+++ b/index.js
@@ -78,7 +78,7 @@ Router.prototype.on = function (method, path, handler, store) {
 
       if (isRegex && (i === len || path.charCodeAt(i) === 47)) {
         nodeType = 3
-      } else if (i < len) {
+      } else if (i < len && path.charCodeAt(i) !== 47) {
         nodeType = 4
       }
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -78,7 +78,7 @@ Router.prototype.on = function (method, path, handler, store) {
 
       if (isRegex && (i === len || path.charCodeAt(i) === 47)) {
         nodeType = 3
-      } else if (i < len) {
+      } else if (i < len && path.charCodeAt(i) !== 47) {
```
brunoscopelliti: The missing of the check against `path.charCodeAt(i)` caused that every parametric part of the url that didn't contain a regular expression, and was followed by another part of the url, was treated as a multi-parametric part.

mcollina:
LGTM

delvedor:
Landed in `v1.6.2` :)

PR #34, opened 2017-10-02
Find refactor

delvedor:
After a lot of work I come up with two solutions:
1. `find2` is a recursive function, we cover all the test cases but we are slower, around 1%-5% leff compared to the iterative version
2. `find` is iterative, we have two trees, one for the static/parametric routes and one for the wildcards. If the static/parametric fails if we know that there is a wildcard route we search it inside the wildcard tree. All the changes inside the `insert` and `_insert` function has been done to support this.

Sorry about the following, two different PRs were a little bit hard to handle:
If you want to test the recursive function:
1. rename `find2`in `find` and viceversa.
2. uncomment line 141

As you can see the recursive version works in all the cases, while the iterative fails in case of nested static nodes. We could fix the iterative version by using a stack, but I'm pretty sure that the performances will slow down a lot.

I'm open to feedbacks and suggestions to improve this, sorry again for the "mess" :P

cc @brunoscopelliti 

ps if you think that two separate pr are better, let me know and I'll working out :P

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,11 +130,11 @@ router.on('GET', '/example/shared/nested/test', (req, res, params) => {
   assert.fail('We should not be here')
 })
 
-router.on('GET', '/example/:param/nested/test', (req, res, params) => {
+router.on('GET', '/example/:param/nested/oops', (req, res, params) => {
   assert.fail('We should not be here')
 })
 
-router.lookup({ method: 'GET', url: '/example/shared/nested/oops' }, res)
+router.lookup({ method: 'GET', url: '/example/shared/nested/oops' }, null)
 ```
 
 <a name="shorthand-methods"></a>
--- a/README.md
+++ b/README.md
@@ -106,6 +106,37 @@ In this case as parameter separator it's possible to use whatever character is n
 
 Having a route with multiple parameters may affect negatively the performance, so prefer single parameter approach whenever possible, especially on routes which are on the hot path of your application.
 
+<a name="match-order"></a>
+##### Match order
+The routes are matched in the following order:
+```
+static
+parametric
+wildcards
+parametric(regex)
+multi parametric(regex)
+```
+This means that if you register two routes, the first static and the second dynamic with a shared part of the path, the static route will always take precedence.
+For example:
+```js
+const assert = require('assert')
+const router = require('find-my-way')({
+  defaultRoute: (req, res) => {
+    assert(req.url === '/example/shared/nested/oops')
+  }
+})
+
+router.on('GET', '/example/shared/nested/test', (req, res, params) => {
+  assert.fail('We should not be here')
+})
+
+router.on('GET', '/example/:param/nested/test', (req, res, params) => {
+  assert.fail('We should not be here')
+})
+
+router.lookup({ method: 'GET', url: '/example/shared/nested/oops' }, res)
+```
+
 <a name="shorthand-methods"></a>
 ##### Shorthand methods
 If you want an even nicer api, you can also use the shorthand methods to declare your routes.
--- a/index.js
+++ b/index.js
@@ -203,6 +203,7 @@ Router.prototype.find = function (method, path) {
   var prefixLen = 0
   var len = 0
   var i = 0
+  var kind = 0
 
   while (true) {
     pathLen = path.length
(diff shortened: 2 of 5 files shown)


brunoscopelliti:
Hey @delvedor 
I’ve took the time to review both the PRs 😉 
Let me list pro/cons of both; feel free to contribute to this list.

**Recursive**
👍  unit tests pass 100% 
👍  do not introduce new structures, and overall it seems simpler

👎  slower; but we could find optimisations.

**Iterative (double tree)**
❔  not sure about performance

👎  failing unit tests

Also a couple of suggestions (please let me know what you think).

**Review benchmark.**
I think that we should review the router tree that is mocked in the benchmark.
IMHO actually it hardly resembles a _typical_ web app routing tree. So I suspect that benchmark optimisations don't necessarily improve throughput of a real server.

For instance I've written your `Node#findByKind` method, so it returns the first node that has the given *kind*, or greater.
The general idea is that when I don't find a node with `kind=2`, I immediately get a node with `kind` 3, or 4 if any, without having to restart the loop over all the children.
However this does not result in any improvements of bench result.

**Wildcard priority.**
Wildcard nodes probably should be verified after parametric with regex nodes (3), and multi-parametric nodes (4); my understanding is that wildcard nodes have node type 2 only cause at the time there were only static (0) and parametric nodes (1).
Since this is a general refactoring we could rework also the priority of each node type.

delvedor:
Hi!
Sorry for delay, I've updated the iterative version and now all the test are passing 🎉 
I've completely removed the recursive version.

Before release this I would like to write some more test and maybe optimize a little more the algorithm.

brunoscopelliti:
I don't know if this is something we want to cover with this refactoring...
The fact that we're now maintaining in memory the last met wildcard node made me think that we may still have problems with parametric nodes.

```
test('Failing test', t => {
  t.plan(1)
  const findMyWay = FindMyWay({
    defaultRoute: (req, res) => {
    }
  })

  findMyWay.on('GET', '/a/completely/static/route', (req, res, params) => {
  })

  findMyWay.on('GET', '/a/:param/static/lol', (req, res, params) => {
    t.equal(req.params, 'completely')
  })

  findMyWay.lookup(
    { method: 'GET', url: 'a/completely/static/lol' },
    null
  )
})

```

mcollina:
@brunoscopelliti I think that case would be hard to support with a prefix-tree. We should probably document it.

delvedor:
@brunoscopelliti I think we should not cover that case, static routes have the precedence over parametric routes and the `completely` path has already been registered as static, so the parameter value could never be `completely`.

brunoscopelliti:
Agree, that it's important to mention this case in the documentation, and a test too.

index.js
```diff
@@ -380,3 +397,20 @@ function fastDecode (path) {
     errored = true
   }
 }
+
+function getWildcardNode (node, method, path, len) {
+  if (node === null) return null
+  var decoded = fastDecode(path.slice(len))
```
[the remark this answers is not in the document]
delvedor: Nice catch!!!! :)

delvedor:
I did some other minor optimization, the benchmarks are slightly better than before.
Please take a final look, if everything is ok for you we can land this.

cc @mcollina @brunoscopelliti

README.md
```diff
@@ -106,6 +106,37 @@ In this case as parameter separator it's possible to use whatever character is n
...
+  defaultRoute: (req, res) => {
+    assert(req.url === '/example/shared/nested/oops')
+  }
+})
+
+router.on('GET', '/example/shared/nested/test', (req, res, params) => {
+  assert.fail('We should not be here')
+})
+
+router.on('GET', '/example/:param/nested/test', (req, res, params) => {
```
mcollina: maybe this should be  `/example/:param/nested/oops`

delvedor: Why?
That example is just to explain the behaviour that is covered by [this](https://github.com/delvedor/find-my-way/blob/find-refactor/test/methods.test.js#L420-L439) test and explained few lines above.

mcollina: I find this example better explaining the concept:

```js
  findMyWay.lookup({ method: 'GET', url: '/example/shared/nested/oops' }, null)
```

brunoscopelliti: I agree with @mcollina , and I would change as well the test that you've linked...
Currently this is different from the example provided here https://github.com/delvedor/find-my-way/pull/34#issuecomment-335070418.

In the current example `/example/shared/nested/oops` is not defined in the router, so there's no much value in checking that we reach the default route.
When we register also an handler for the route `/example/shared/nested/oops`, we still expect to reach the  default route, because the initial part of tha parametric route is shadowed by the static `/example/shared/nested/test` route.
I guess this was also Matteo's point.

test/issue-28.test.js
```diff
@@ -499,3 +499,79 @@ test('Nested wildcards with parametric and static - 6', t => {
...
+
+  findMyWay.on('GET', '/foo3/:param', (req, res, params) => {
+    t.is(req.url, '/foo3/hello')
+  })
+
+  findMyWay.on('GET', '/foo3/*', (req, res, params) => {
+    t.fail('we should not be here, the url is: ' + req.url)
+  })
+
+  findMyWay.on('GET', '/foo4/param/hello/test/long/route', (req, res, params) => {
```
brunoscopelliti: I find it hard reading a route like this, cause I've to figure out from the context of the test whether `param` was really thought as a static part of the route, or maybe it was meant to be a parameter, and there's a typo.
So I suggest to not use the word `param` in a long static route... we can use whatever else.

This occurs also in a couple of other tests in this file.

delvedor:
@brunoscopelliti any other feedback? :)

brunoscopelliti:
@delvedor 
Great work!

Issue #40, opened 2017-11-07
Support for .off method to unmount a route


hekike:
Would be nice to have an `off` method on the router instance to be able to unmount already attached handler.

```js
const router = require('find-my-way')()

function handler (req, res, params) {
  res.end('{"message":"hello world"}')
}

router.on('GET', '/', handler)
router.off('GET', '/') // or router.off('GET', '/', handler)
```

delvedor:
Closed by #43

PR #43, opened 2017-11-17
feat(off): add off feature

hekike:
Add `.off(method, path)` method.

## Related issues

- https://github.com/delvedor/find-my-way/issues/40

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,31 @@ parametric(regex)
 multi parametric(regex)
 ```
 
+<a name="off"></a>
+#### off(method, path)
+Deregister a route.
+```js
+router.off('GET', '/example')
+// => { handler: Function, params: Object, store: Object}
+// => null
+```
+
+##### off(methods[], path, handler, [store])
+Deregister a route for each method specified in the `methods` array.
+It comes handy when you need to deregister multiple routes with the same path but different methods.
+```js
+router.off(['GET', 'POST'], '/example')
+// => [{ handler: Function, params: Object, store: Object}]
+// => null
+```
+
+<a name="reset"></a>
+#### reset()
+Empty router.
+```js
+router.reset()
+```
+
 ##### Caveats
 * Since *static* routes have greater priority than *parametric* routes, when you register a static route and a dynamic route, which have part of their path equal, the static route shadows the parametric route, that becomes not accessible. For example:
 ```js
--- a/index.js
+++ b/index.js
@@ -35,6 +35,7 @@ function Router (opts) {
   }
 
   this.tree = new Node()
+  this.routes = []
 }
 
 Router.prototype.on = function on (method, path, handler, store) {
@@ -57,6 +58,13 @@ Router.prototype.on = function on (method, path, handler, store) {
 
   const params = []
   var j = 0
+
+  this.routes.push({
+    method: method,
+    path: path,
+    handler: handler,
+    store: store
+  })
 
   for (var i = 0, len = path.length; i < len; i++) {
     // search for parametric or wildcard routes
@@ -181,6 +189,38 @@ Router.prototype._insert = function _insert (method, path, kind, params, handler
     }
     return
   }
+}
+
+Router.prototype.reset = function reset () {
+  this.tree = new Node()
+  this.routes = []
+}
+
+Router.prototype.off = function off (method, path) {
+  var self = this
+
+  if (Array.isArray(method)) {
+    return method.map(function (method) {
+      return self.off(method, path)
+    })
+  }
+
+  // method validation
+  assert(typeof method === 'string', 'Method should be a string')
+  assert(httpMethods.indexOf(method) !== -1, `Method '${method}' is not an http method.`)
+  // path validation
+  assert(typeof path === 'string', 'Path should be a string')
+  assert(path.length > 0, 'The path could not be empty')
+  assert(path[0] === '/' || path[0] === '*', 'The first character of a path should be `/` or `*`')
+
+  // Rebuild tree without the specific route
+  var newRoutes = self.routes.filter(function (route) {
+    return !(method === route.method && path === route.path)
+  })
+  self.reset()
+  newRoutes.forEach(function (route) {
+    self.on(route.method, route.path, route.handler, route.store)
+  })
 }
 
 Router.prototype.lookup = function lookup (req, res) {
--- a/test/methods.test.js
+++ b/test/methods.test.js
@@ -5,11 +5,12 @@ const test = t.test
 const FindMyWay = require('../')
 
 test('the router is an object with methods', t => {
-  t.plan(3)
+  t.plan(4)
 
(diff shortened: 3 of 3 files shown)


delvedor:
Hi! Thank you for the proposal!

I think the best option would be regenerate the entire tree every time you call `.off`.
1. No need to change the internals which are highly optimized.
2. add the parent node to a Node means that you are creating a circular reference, in this way you can't print the tree (`JSON.stringify(router.tree, null, 2)`) which is very useful for debugging.

We can create an array where we save all the routes and if the user calls `.off` we remove the specific route from that array and then we regenerate the tree.
What do you think?

hekike:
@delvedor thanks for the feedback. I'm okay with the solution what you proposed.
Do you want to work on it or should I go ahead?

delvedor:
@hekike go ahead! :)

hekike:
@delvedor done, how about now? ;)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "find-my-way",
-  "version": "1.7.0",
+  "version": "1.7.1",
```
mcollina: Why this?

index.js
```diff
@@ -183,6 +190,38 @@ Router.prototype._insert = function _insert (method, path, kind, params, handler
...
+    })
+  }
+
+  // method validation
+  assert(typeof method === 'string', 'Method should be a string')
+  assert(httpMethods.indexOf(method) !== -1, `Method '${method}' is not an http method.`)
+  // path validation
+  assert(typeof path === 'string', 'Path should be a string')
+  assert(path.length > 0, 'The path could not be empty')
+  assert(path[0] === '/' || path[0] === '*', 'The first character of a path should be `/` or `*`')
```
mcollina: it seems a part of this check is not really tested.

README.md
```diff
@@ -117,6 +117,31 @@ parametric(regex)
...
+// => [{ handler: Function, params: Object, store: Object}]
+// => null
+```
+
+<a name="reset"></a>
+#### reset()
+Empty router.
+```js
+router.reset()
+```
```
mcollina: can you add a test for reset()?

delvedor:
Try add this test:
```js
test('off with nested wildcards with parametric and static', t => {
  t.plan(2)
  const findMyWay = FindMyWay({
    defaultRoute: (req, res) => {
    }
  })

    t.is(params['*'], '/foo2/first/second')
  })
  findMyWay.on('GET', '/foo1/*', () => {})
  findMyWay.on('GET', '/foo2/*', () => {})
  findMyWay.on('GET', '/foo3/:param', () => {})
  findMyWay.on('GET', '/foo3/*', () => {})
  findMyWay.on('GET', '/foo4/param/hello/test/long/route', () => {})

  findMyWay.off('GET', '/foo3/*')
  t.notOk(findMyWay.find('GET', '/foo3/first/second'))

  findMyWay.off('GET', '/foo2/*')
  findMyWay.lookup(
    { method: 'GET', url: '/foo2/first/second' },
    null
  )
})
```
For some reason is not passing on my machine.

hekike:
@delvedor are you sure it's related to the `.off`, for me, it's failing any `off` call.

```js
test('off with nested wildcards with parametric and static', t => {
  t.plan(1)
  const findMyWay = FindMyWay({
    defaultRoute: (req, res) => {
    }
  })

    t.is(params['*'], '/foo2/first/second')
  })
  findMyWay.on('GET', '/foo1/*', () => {})
  findMyWay.on('GET', '/foo2/*', () => {})
  findMyWay.on('GET', '/foo3/:param', () => {})
  findMyWay.on('GET', '/foo3/*', () => {})
  findMyWay.on('GET', '/foo4/param/hello/test/long/route', () => {})

  // findMyWay.off('GET', '/foo3/*')
  // t.notOk(findMyWay.find('GET', '/foo3/first/second'))

  // findMyWay.off('GET', '/foo2/*')
  findMyWay.lookup(
    { method: 'GET', url: '/foo2/first/second' },
    null
  )
})
```

delvedor:
@hekike this is the standard situation:
```js
test('check 1', t => {
  t.plan(1)
  const findMyWay = FindMyWay({
    defaultRoute: (req, res) => {
    }
  })

  findMyWay.on('GET', '*', () => {})
  findMyWay.on('GET', '/foo1/*', () => {})
  findMyWay.on('GET', '/foo2/*', (req, res, params) => {
    t.is(params['*'], 'first/second')
  })
  findMyWay.on('GET', '/foo3/:param', () => {})
  findMyWay.on('GET', '/foo3/*', () => {})
  findMyWay.on('GET', '/foo4/param/hello/test/long/route', () => {})

  findMyWay.lookup(
    { method: 'GET', url: '/foo2/first/second' },
    null
  )
})
```

And this is the one without `/foo2/*`:
```js
test('check 2', t => {
  t.plan(1)
  const findMyWay = FindMyWay({
    defaultRoute: (req, res) => {
    }
  })

    t.is(params['*'], '/foo2/first/second')
  })
  findMyWay.on('GET', '/foo1/*', () => {})
  findMyWay.on('GET', '/foo3/:param', () => {})
  findMyWay.on('GET', '/foo3/*', () => {})
  findMyWay.on('GET', '/foo4/param/hello/test/long/route', () => {})

  findMyWay.lookup(
    { method: 'GET', url: '/foo2/first/second' },
    null
  )
})
```
Both tests are passing on my machine, this means that `off` is messing up something :S

hekike:
@delvedor I found the issue and fixed it. Thanks for the help!

hekike:
@delvedor I also added your wildcard test case to the tests.

delvedor:
@hekike you are welcome!
For the sake of curiosity, what did you changed?

test/methods.test.js
```diff
@@ -37,6 +67,124 @@ test('register a route with multiple methods', t => {
...
+    t.is(params['*'], '/foo2/first/second')
+  })
+  findMyWay.on('GET', '/foo1/*', () => {})
+  findMyWay.on('GET', '/foo2/*', () => {})
+  findMyWay.on('GET', '/foo3/:param', () => {})
+  findMyWay.on('GET', '/foo3/*', () => {})
+  findMyWay.on('GET', '/foo4/param/hello/test/long/route', () => {})
+
+  var route1 = findMyWay.find('GET', '/foo2/first/second')
+  t.ok(route1.params['*'])
```
delvedor: Can you check also the value of the wildcard?
eg:
```js
t.is(route1.params['*'], 'first/second')
```

Same below :)

hekike:
@delvedor I added the param checks to the tests.
The issue was earlier that I pushed the route to late some return run before.

delvedor:
Landed in [`v1.8.0`](https://github.com/delvedor/find-my-way/releases/tag/v1.8.0) :)

hekike:
@delvedor thanks! 🎉

Issue #33, opened 2017-10-02
AssertionError: Method 'GET' already declared for route


brunoscopelliti:
I came across this (still failing) test case while thinking a general solution for the issue https://github.com/delvedor/find-my-way/pull/29#issuecomment-333401951

```js
test('Similar route path, different parameter', t => {
  t.plan(2)
  const findMyWay = FindMyWay({
    defaultRoute: (req, res) => {
      t.fail('we should not be here, the url is: ' + req.url)
    }
  })
  findMyWay.on('GET', '/user/:userId(^\\d+)/orders', (req, res, params) => {
    t.equal(params.userId, '42')
  })
  findMyWay.on('GET', '/user/:username(^[a-z]+)/orders', (req, res, params) => {
    t.equal(params.username, 'foo')
  })
  findMyWay.lookup({ method: 'GET', url: '/user/42/orders' }, null)
  findMyWay.lookup({ method: 'GET', url: '/user/foo/orders' }, null)
})
```

Currently it's not possible to configure a routing such as the one above:
_AssertionError: Method 'GET' already declared for route '/orders'_.

Do you think find-my-way should support this use case?

delvedor:
Yes, this is correct, you have already defined a node of kind 3.

We could support that kind of behavior, but I'm pretty sure that the performances will slow down a lot and the overall complexity will increase.
@mcollina thoughts?

mcollina:
IMHO we should not support it, but provide a little bit more guidance in how to handle the situation. The current error is not explaining the what is happening and how to fix it.

brunoscopelliti:
> but provide a little bit more guidance in how to handle the situation

@mcollina could you please clarify what you mean...
Do you intend to mention this case in the documentation, or to change the code to make the error more descriptive? (or both perhaps?)

delvedor:
Probably both, but detect that error is not super easy, it will require a custom logic to "see what is around".

Hume3:
Why did you decide to use listeners instead of simple methods or functions ?, These are really problematic, I made an adaptation for microJS and I had difucltades, to use hot module reload. I had to add the events .on () and remove them .off (), but it really looks a bit ambiguous and I do not feel trust, I think the syntax used by express router or other routers they are more solids.

Find-my-way: 
Speed -> Relly good :D
Implentation -> Very bad :(

it's just a constructive criticism.

Greetings.

delvedor:
> Why did you decide to use listeners instead of simple methods or functions ?

[...]

PR #45, opened 2017-12-03
Fixes 44

delvedor:
As titled.

### DIFF
--- a/index.js
+++ b/index.js
@@ -244,6 +244,7 @@ Router.prototype.find = function find (method, path) {
     var prefix = currentNode.prefix
     var prefixLen = prefix.length
     var len = 0
+    var previousPath = path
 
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -283,7 +284,13 @@ Router.prototype.find = function find (method, path) {
 
     var node = currentNode.find(path[0], method)
     if (node === null) {
-      return getWildcardNode(wildcardNode, method, originalPath, pathLenWildcard)
+      node = currentNode.parametricBrother
```
mcollina: can this have two parametricBrothers?

delvedor: At the moment is not supported, we have already discussed that in https://github.com/delvedor/find-my-way/issues/33 :)

PR #47, opened 2017-12-21
Fixes 46

delvedor:
As titled.
The problem was that we were saving the wildcard node even if there was no matching node, just by moving the wildcard child check the issue is fixed.

@nwoltman can you confirm that this fixes your issue?


nwoltman:
@delvedor Yes this fixes my issue :)

PR #50, opened 2018-01-18
Add support for ignoring trailing slashes

jsumners:
This PR implements the logic described in https://github.com/fastify/fastify/issues/584#issuecomment-355134797

```js
const router = require('find-my-way')({
  trimTrailingSlash: true
})

function handler (req, res, params) {
  res.send('foo')
}

// maps "/foo/" and "/foo" to `handler`
router.on('GET', '/foo/', handler)
```

### DIFF
--- a/test/server.test.js
+++ b/test/server.test.js
@@ -112,3 +112,113 @@ test('automatic default route', t => {
     })
   })
 })
+
+test('maps two routes when trailing slash should be trimmed', t => {
+  t.plan(25)
+  const findMyWay = FindMyWay({
+    trimTrailingSlash: true
+  })
+
+  findMyWay.on('GET', '/test/', (req, res, params) => {
+    t.ok(req)
+    t.ok(res)
+    t.ok(params)
+    res.end('test')
+  })
+
+  findMyWay.on('GET', '/othertest', (req, res, params) => {
+    t.ok(req)
+    t.ok(res)
+    t.ok(params)
+    res.end('othertest')
+  })
+
+  const server = http.createServer((req, res) => {
+    findMyWay.lookup(req, res)
+  })
+
+  server.listen(0, err => {
+    t.error(err)
+    server.unref()
+
+    const baseURL = 'http://localhost:' + server.address().port
+
+    request({
+      method: 'GET',
+      uri: baseURL + '/test/'
+    }, (err, response, body) => {
+      t.error(err)
+      t.strictEqual(response.statusCode, 200)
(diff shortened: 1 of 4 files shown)


jsumners:
Sorry for all of the incoming coveralls messages. I had to do a few force pushes to fix some README items.

test/server.test.js
```diff
@@ -112,3 +112,46 @@ test('automatic default route', t => {
...
   })
 })
+
+test('maps two routes when trailing slash should be trimmed', t => {
+  t.plan(13)
+  const findMyWay = FindMyWay({
+    trimTrailingSlash: true
+  })
+
+  findMyWay.on('GET', '/test/', (req, res, params) => {
```
delvedor: Can you also add another route where you register `/othertest` (without the trailing slash) and test how it behaves?
Then it could be useful add a different test with `trimTrailingSlash` set to `false` and test how it behaves as well. :)

jsumners:
As I said in https://github.com/fastify/fastify/issues/584#issuecomment-354607698 I do not think it would be wise to default this option to `true`.

.vscode/launch.json
```diff
@@ -0,0 +1,14 @@
+{
```
delvedor: Can you remove this file? :D

jsumners: It's gone in the next commit. Accidentally did a `git add .`.

jsumners:
Hang on. Thought of one more scenario.

jsumners:
Okay, give this a thorough review. It's possible I missed something in the `.off` logic.

nwoltman:
Looking great @jsumners! One thing I've noticed though is that the option name is a little unclear, since `router.on('GET', '/path')` will also register `router.on('GET', '/path/')`.

[`path-to-regexp`](https://github.com/pillarjs/path-to-regexp#usage) has a similar option that they call `"strict"`. Maybe we could call it something like `"strictTrailingSlash"` here (or `"optionalTrailingSlash"` so you don't have to go and reverse all of the boolean logic).

jsumners:
@nwoltman I have renamed it to `ignoreTrailingSlash`.

nwoltman:
@jsumners Sounds great 👍

nwoltman:
A couple more cases to handle:

+ If the path ends with a wildcard `*`, it's not necessary to add a trailing slash
+ If the path is exactly `'/'`, just register that path (avoid trimming it to the empty string `''`)

jsumners:
`/` and `*` cases covered.

PR #51, opened 2018-01-20
Use constants for node types instead of magic numbers

jsumners:
This PR replaces all of the instances like `node.kind === 0` with `node.kind === NODE_TYPES.STATIC`. This is clearer to the reader, and thus to people trying to contribute to the project.

My system is a mid-2012 MacBook Pro, so it isn't very fast. But I don't see any real discrepancy in the benchmarks:

**master:**
```
lookup static route x 22,606,815 ops/sec ±1.10% (85 runs sampled)
lookup dynamic route x 1,406,844 ops/sec ±0.99% (89 runs sampled)
lookup dynamic multi-parametric route x 776,795 ops/sec ±0.99% (87 runs sampled)
lookup dynamic multi-parametric route with regex x 689,269 ops/sec ±0.99% (89 runs sampled)
lookup long static route x 2,919,538 ops/sec ±1.03% (89 runs sampled)
find static route x 27,170,227 ops/sec ±0.89% (87 runs sampled)
find dynamic route x 1,490,351 ops/sec ±1.77% (89 runs sampled)
find dynamic multi-parametric route x 832,010 ops/sec ±0.89% (91 runs sampled)
find dynamic multi-parametric route with regex x 715,763 ops/sec ±1.00% (87 runs sampled)
find long static route x 3,852,285 ops/sec ±0.82% (91 runs sampled)
```
**branch:**
```
lookup static route x 22,736,075 ops/sec ±1.77% (85 runs sampled)
lookup dynamic route x 1,415,499 ops/sec ±0.90% (91 runs sampled)
lookup dynamic multi-parametric route x 733,222 ops/sec ±1.10% (87 runs sampled)
lookup dynamic multi-parametric route with regex x 701,346 ops/sec ±0.96% (91 runs sampled)
lookup long static route x 2,912,104 ops/sec ±1.06% (91 runs sampled)
find static route x 26,740,260 ops/sec ±1.01% (86 runs sampled)
find dynamic route x 1,489,647 ops/sec ±1.15% (89 runs sampled)
find dynamic multi-parametric route x 829,876 ops/sec ±0.96% (88 runs sampled)
find dynamic multi-parametric route with regex x 736,822 ops/sec ±0.92% (90 runs sampled)
find long static route x 3,729,310 ops/sec ±1.11% (84 runs sampled)
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,14 +1,6 @@
 'use strict'
 
 /*
-  Node type
-    static: 0,
-    param: 1,
-    matchAll: 2,
-    regex: 3
-    multi-param: 4
-      It's used for a parameter, that is followed by another parameter in the same part
-
   Char codes:
     '#': 35
     '*': 42
@@ -20,6 +12,7 @@
 
 const assert = require('assert')
 const Node = require('./node')
+const NODE_TYPES = Node.prototype.types
 const httpMethods = ['DELETE', 'GET', 'HEAD', 'PATCH', 'POST', 'PUT', 'OPTIONS', 'TRACE', 'CONNECT']
 var errored = false
 
@@ -86,7 +79,7 @@ Router.prototype._on = function _on (method, path, handler, store) {
     // search for parametric or wildcard routes
     // parametric route
     if (path.charCodeAt(i) === 58) {
-      var nodeType = 1
+      var nodeType = NODE_TYPES.PARAM
       j = i + 1
       this._insert(method, path.slice(0, i), 0, null, null, null, null)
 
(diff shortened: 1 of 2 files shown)


jsumners:
@delvedor why? I think `types.STATIC` is clearer than `STATIC_TYPE` or `TYPE_STATIC`.

jsumners:
Also, I don't see any way to implement them as standalone variables without duplicating the code amongst both the `node.js` and `index.js` files. At least, not while the module supports Node v4 since that doesn't support destructuring.

delvedor:
I guess that you used Node v8 to run you benchmarks, do you get the same result with Node v6?
When I started developing this router I was using Node v6 and I remember I ran different tests to find the best solution. If I remember correctly Cranckshaft didn't perform this optimization, while TurboFan does it.
Since Node v6 is still a thing I think we should remain with this configuration a little longer.

jsumners:
Looks equivalent to me:

find-my-way zsh

jsumners:
Should I close this?

PR #54, opened 2018-01-27
Added maxParamLength option

delvedor:
As titled.

### DIFF
--- a/index.js
+++ b/index.js
@@ -35,6 +35,7 @@ function Router (opts) {
   }
 
   this.ignoreTrailingSlash = opts.ignoreTrailingSlash || false
+  this.maxParamLength = opts.maxParamLength || 100
   this.tree = new Node()
   this.routes = []
 }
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -347,7 +349,9 @@ Router.prototype.find = function find (method, path) {
     if (kind === 1) {
       currentNode = node
       i = 0
-      while (i < pathLen && path.charCodeAt(i) !== 47) i++
+      while (i < pathLen && path.charCodeAt(i) !== 47) {
+        if (i++ > maxParamLength) return null
+      }
```
mcollina: I think you can move the if check after the while loop. In this way we will not be checking it on every increment, and we should not have a perf regression.

delvedor: I thought about that, but if we check that value continuously the user will be safe against DoS attacks, because since Node we can perform only one operation at the time I don't want to waste time searching for the next slash if the path is extremely long.
What do you think?

mcollina: I think the time to do that iteration is so fast anyway that it does not matter. There is a node limit anyway, this should be at max 1024 (if I remember correctly). You should benchmark it and see the difference

delvedor: Move the if check outside the while hook (obiviously) saves some perf.

PR #57, opened 2018-02-01
Fixes #49

### LINKED ISSUE #49: Defining static route after parametric route causes issue
Issue based on fastify/fastify#573

Works:

```js

router.on('GET', '/static', function s () {}) // Static route defined first
router.on('GET', '/:param', function p () {})

router.find('GET', '/static')
// { handler: [Function: s], params: {}, store: null }

router.find('GET', '/para')
// { handler: [Function: p], params: { param: 'para' }, store: null }

router.find('GET', '/s')
// { handler: [Function: p], params: { param: 's' }, store: null }
```

Does not work:

```js

router.on('GET', '/:param', function p () {})
router.on('GET', '/static', function s () {}) // Static route defined second

router.find('GET', '/static')
// { handler: [Function: s], params: {}, store: null }

router.find('GET', '/para')
// { handler: [Function: p], params: { param: 'para' }, store: null }

router.find('GET', '/s')
// null
```

Expected `router.find('GET', '/s')` to return the same object as above.

Related to #44 which was fixed (except for this issue) by #45.

delvedor:
As titled.
cc @nwoltman @jsumners

### DIFF
--- a/node.js
+++ b/node.js
@@ -38,7 +38,7 @@ Node.prototype.add = function (node) {
   var parametricBrother = null
   for (var i = 0; i < this.numberOfChildren; i++) {
     const child = this.children[i]
-    if ([this.types.PARAM, this.types.REGEX, this.types.MULTI_PARAM].indexOf(node.kind) > -1) {
+    if ([this.types.PARAM, this.types.REGEX, this.types.MULTI_PARAM].indexOf(child.kind) > -1) {
       parametricBrother = child
       break
     }
--- a/node.js
+++ b/node.js
@@ -27,17 +27,25 @@ Node.prototype.add = function (node) {
     this.wildcardChild = node
   }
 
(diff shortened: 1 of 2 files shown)


node.js
```diff
@@ -30,17 +30,25 @@ Node.prototype.add = function (node) {
...
-
   this.children.push(node)
   this.children.sort((n1, n2) => n1.kind - n2.kind)
   this.numberOfChildren++
+
+  // Search for a parametric brother and store it in a variable
+  for (var i = 0; i < this.numberOfChildren; i++) {
+    const child = this.children[i]
+    if ([this.types.PARAM, this.types.REGEX, this.types.MULTI_PARAM].indexOf(node.kind) > -1) {
+      var parametricBrother = child
```
mcollina: can you move `var parametricBrother` to the top?

Issue #60, opened 2018-02-14
Found a wrong route


allevo:
The following code finds a wrong route
```js
const noop = () => {}
const findMyWay = FindMyWay()

findMyWay.on('GET', '/bb/', noop)
findMyWay.on('GET', '/bb/bulk', noop)

const f = findMyWay.find('GET', '/bulk')

// f === noop
```

Expected: 
`f` should be null

Issue born here: https://github.com/fastify/fastify/pull/787

delvedor:
That's strange, if you remove the trailing slash in the first route the test passes.
```js
'use strict'

const t = require('tap')
const test = t.test
const FindMyWay = require('../')
const noop = () => {}

test('issue-60', t => {
  t.plan(1)
  const findMyWay = FindMyWay()

  findMyWay.on('GET', '/bb', noop)
  findMyWay.on('GET', '/bb/bulk', noop)

  t.equal(findMyWay.find('GET', '/bulk'), null)
})
```

cc @nwoltman

delvedor:
It passes also if you use the `ignoreTrailingSlash` option.
```js
'use strict'

const t = require('tap')
const test = t.test
const FindMyWay = require('../')
const noop = () => {}

test('issue-60', t => {
  t.plan(1)
  const findMyWay = FindMyWay({ ignoreTrailingSlash: true })

  findMyWay.on('GET', '/bb/', noop)
  findMyWay.on('GET', '/bb/bulk', noop)

  t.equal(findMyWay.find('GET', '/bulk'), null)
})
```

nwoltman:
I'm guessing it has something to do with how lookups happen. The tree structure might give us some information on why this is happening:

## 1. `/b/bulk`

```js
var router = require('find-my-way')()
router.on('GET', '/b/bulk', function b_bulk () {})
console.log(router.prettyPrint())
```

```
└── /
    └── b/bulk (GET)
```

## 2. `/b` and `/b/bulk`

```js
var router = require('find-my-way')()
router.on('GET', '/b', function b () {})
router.on('GET', '/b/bulk', function b_bulk () {})
console.log(router.prettyPrint())
```

```
└── /
    └── b (GET)
        └── /bulk (GET)
```

## 3. `/b/` and `/b/bulk` (Problem)

```js
var router = require('find-my-way')()
router.on('GET', '/b/', function b_ () {})
router.on('GET', '/b/bulk', function b_bulk () {})
console.log(router.prettyPrint())
```

```
└── /
[...]

PR #61, opened 2018-02-14
Show #60

allevo:
Add test for showing the #60

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,6 +16,8 @@ const NODE_TYPES = Node.prototype.types
 const httpMethods = ['DELETE', 'GET', 'HEAD', 'PATCH', 'POST', 'PUT', 'OPTIONS', 'TRACE', 'CONNECT']
 var errored = false
 
+const ROUTE_NOT_FOUND = null
+
 function Router (opts) {
   if (!(this instanceof Router)) {
     return new Router(opts)
@@ -253,7 +255,7 @@ Router.prototype.off = function off (method, path) {
 
 Router.prototype.lookup = function lookup (req, res) {
   var handle = this.find(req.method, sanitizeUrl(req.url))
-  if (!handle) return this._defaultRoute(req, res)
+  if (handle === ROUTE_NOT_FOUND) return this._defaultRoute(req, res)
   return handle.handler(req, res, handle.params, handle.store)
 }
 
@@ -290,7 +292,7 @@ Router.prototype.find = function find (method, path) {
           }
         }
         if (isStatic && ((previousPrefix + previousPath) !== originalPath)) {
-          return null
+          return ROUTE_NOT_FOUND
         }
         return {
           handler: handle.handler,
@@ -403,7 +405,7 @@ Router.prototype.find = function find (method, path) {
       }
       decoded = fastDecode(path.slice(0, i))
       if (errored) {
-        return null
+        return ROUTE_NOT_FOUND
       }
       params[pindex++] = decoded
       path = path.slice(i)
@@ -489,10 +491,10 @@ function fastDecode (path) {
 }
 
 function getWildcardNode (node, method, path, len) {
-  if (node === null) return null
+  if (node === null) return ROUTE_NOT_FOUND
   var decoded = fastDecode(path.slice(-len))
   if (errored) {
-    return null
+    return ROUTE_NOT_FOUND
   }
   var handle = node.getHandler(method)
   if (handle !== undefined) {
@@ -502,5 +504,5 @@ function getWildcardNode (node, method, path, len) {
       store: handle.store
     }
   }
-  return null
+  return ROUTE_NOT_FOUND
 }
--- a/index.js
+++ b/index.js
@@ -291,7 +291,7 @@ Router.prototype.find = function find (method, path) {
             paramsObj[paramNames[i]] = params[i]
           }
         }
-        if (isStatic && ((previousPrefix + previousPath) !== originalPath)) {
+        if (isStatic === true && ((previousPrefix + previousPath) !== originalPath)) {
           return ROUTE_NOT_FOUND
         }
         return {
@@ -323,7 +323,7 @@ Router.prototype.find = function find (method, path) {
     }
     var kind = node.kind
 
-    previousPrefix = previousPrefix + currentNode.prefix
+    previousPrefix += currentNode.prefix
 
     // static route
     if (kind === NODE_TYPES.STATIC) {
--- a/index.js
+++ b/index.js
@@ -267,6 +267,8 @@ Router.prototype.find = function find (method, path) {
   var pindex = 0
   var params = []
   var i = 0
+  var previousPrefix = ''
+  var isStatic = true
 
   while (true) {
     var pathLen = path.length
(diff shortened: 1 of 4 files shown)


mcollina:
is there any update on this? Would you have time to contribute a fix as well?

allevo:
I'm working on this, but it is not simple. yesterday evening I may find the problem (the code forgets the previous prefix) but I need time to implement it. After that I'll update the PR asap.

If anyone has any suggestions, please come forward!

allevo:
@mcollina @delvedor fixed. Could you review this?

index.js
```diff
@@ -287,7 +291,9 @@ Router.prototype.find = function find (method, path) {
             paramsObj[paramNames[i]] = params[i]
           }
         }
-
+        if (isStatic && ((previousPrefix + previousPath) !== originalPath)) {
```
delvedor: Can you use `isStatic === true`?

index.js
```diff
@@ -317,6 +323,8 @@ Router.prototype.find = function find (method, path) {
     }
     var kind = node.kind
 
+    previousPrefix = previousPrefix + currentNode.prefix
```
delvedor: I think here you can do `previousPrefix += prefix`

nwoltman:
Yes looks like this fixes the problem 👍

allevo:
```
$ git checkout master 
Switched to branch 'master'
$ npm run bench

> find-my-way@1.10.1 bench /Users/allevo/repos/find-my-way
> node bench.js

lookup static route x 16,146,561 ops/sec ±1.94% (81 runs sampled)
lookup dynamic route x 1,102,008 ops/sec ±1.75% (85 runs sampled)
lookup dynamic multi-parametric route x 605,942 ops/sec ±2.16% (85 runs sampled)
lookup dynamic multi-parametric route with regex x 605,571 ops/sec ±1.16% (89 runs sampled)
lookup long static route x 2,588,693 ops/sec ±1.52% (89 runs sampled)
find static route x 17,640,058 ops/sec ±2.59% (82 runs sampled)
find dynamic route x 1,162,933 ops/sec ±3.05% (86 runs sampled)
find dynamic multi-parametric route x 642,138 ops/sec ±2.15% (87 runs sampled)
find dynamic multi-parametric route with regex x 635,571 ops/sec ±1.21% (90 runs sampled)
find long static route x 3,183,752 ops/sec ±1.75% (85 runs sampled)
$ git checkout show-60 
Switched to branch 'show-60'
$ npm run bench

> find-my-way@1.10.1 bench /Users/allevo/repos/find-my-way
> node bench.js

lookup static route x 15,783,816 ops/sec ±1.78% (80 runs sampled)
lookup dynamic route x 1,055,399 ops/sec ±1.67% (85 runs sampled)
lookup dynamic multi-parametric route x 568,765 ops/sec ±2.07% (84 runs sampled)
lookup dynamic multi-parametric route with regex x 532,505 ops/sec ±3.51% (83 runs sampled)
lookup long static route x 1,535,436 ops/sec ±2.79% (80 runs sampled)
find static route x 15,120,369 ops/sec ±1.65% (80 runs sampled)
find dynamic route x 1,083,889 ops/sec ±1.91% (82 runs sampled)
find dynamic multi-parametric route x 580,661 ops/sec ±1.95% (82 runs sampled)
find dynamic multi-parametric route with regex x 541,456 ops/sec ±1.65% (82 runs sampled)
find long static route x 1,879,066 ops/sec ±1.68% (84 runs sampled)
```

Not so good.

IMHO I can avoid to concatenate the prefix storing the absolutePrefix inside the node

index.js
```diff
@@ -267,6 +269,8 @@ Router.prototype.find = function find (method, path) {
   var pindex = 0
   var params = []
   var i = 0
+  var previousPrefix = ''
+  var isStatic = true
```
mcollina: I think `isStatic`  should be reset to true at some point in the logic. If not we are not fixing this bug in its most complete form, as `/static/:parametric/static/:parametric` should be supported as well

allevo:
> can you please restore null? I fear there might be some form of context escaping that might prevent some optimizations.

the restore changes nothing

index.js
allevo: me too but the the following test passes with this implementation:
```js

  t.test('with parameter / 5', t => {
    t.plan(2)
    const findMyWay = FindMyWay()

    findMyWay.on('GET', '/bb/:foo/aa/', noop)
    findMyWay.on('GET', '/bb/:foo/aa/bulk', noop)

    t.equal(findMyWay.find('GET', '/bulk'), null)
    t.equal(findMyWay.find('GET', '/bb/foo/bulk'), null)
  })

```

nwoltman: That's not the same as what @mcollina said. That test only has one `:parametric` part (`:foo`).

allevo: I'm bit lost. The following test passes too:
```js

  t.test('with parameter / 6', t => {
    t.plan(3)
    const findMyWay = FindMyWay()

    findMyWay.on('GET', '/static/:parametric/static/:parametric', noop)
    findMyWay.on('GET', '/static/:parametric/static/:parametric/bulk', noop)

    t.equal(findMyWay.find('GET', '/bulk'), null)
    t.equal(findMyWay.find('GET', '/static/foo/bulk'), null)
    t.notEqual(findMyWay.find('GET', '/static/foo/static/bulk'), null)
  })

```

Let me know which test I should add

mcollina: the `isStatic` flag switches only once and it is not reset. I think it should be.

allevo: Me too. But I don't know when and where. Without a failing test is very difficult for me to reset that flag.

mcollina:
Go ahead and add the absolutePrefix thing

mcollina:
> I suppose the performance deterioration is for the string concatenation

The performance deterioration is due to the addition string comparison check (as far I read the code). Maybe it's something else.

index.js
delvedor: What does it mean `isStatic`?
The current node is static? The path is  static?

index.js
```diff
@@ -317,23 +323,27 @@ Router.prototype.find = function find (method, path) {
     }
     var kind = node.kind
 
+    previousPrefix += currentNode.prefix
```
delvedor: Here you can directly use `prefix`, it is declared at the beginning of the loop.

index.js
allevo: All the node chain (node, its parent and so on) is static. This bug happens only on static route. I'm not able to reproduce it with the url with parameters

index.js
```diff
@@ -306,9 +312,9 @@ Router.prototype.find = function find (method, path) {
     }
 
     var node = currentNode.find(path[0], method)
-    if (node === null) {
+    if (node === ROUTE_NOT_FOUND) {
       node = currentNode.parametricBrother
-      if (node === null) {
+      if (node === ROUTE_NOT_FOUND) {
```
nwoltman: `null` is actually easier to understand here. `ROUTE_NOT_FOUND` sounds like the entire route was not found so the algorithm should return immediately, but here all it means is that it didn't match one of the node types. I'd really suggest reverting all the `ROUTE_NOT_FOUND` changes. `null` is perfectly easy to understand.

index.js
```diff
@@ -306,9 +312,9 @@ Router.prototype.find = function find (method, path) {
     }
 
     var node = currentNode.find(path[0], method)
```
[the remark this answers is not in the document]
delvedor: I think that part of the problem is the node `label` property. We should investigate also there.

delvedor:
Closed by #65.
Thank you anyway!

PR #64, opened 2018-02-27
Invalid expressions for regexp parameters

### LINKED ISSUE #62: Regex groups seem to break it
Had a  regex with a group inside which breaks the parsing `/foo/:id(([a-f0-9]{32},?)+)`

``` js
var find = require('find-my-way')()

find.get('/foo/:id(([a-f0-9]{32},?)+)', function () {})
```

### LINKED ISSUE #63: Unclosed regex spins the cpu for ever
The following unclosed regex makes find-my-way spin the cpu forever. Throwing an exception could be nicer :)

``` js
var find = require('find-my-way')()

find.get('/foo/:id(a', function () {})
```

absolux:
Fixes #63 by throwing a `TypeError` error if the regexp expressions is wrong
Fixes #62

### DIFF
--- a/index.js
+++ b/index.js
@@ -88,7 +88,9 @@ Router.prototype._on = function _on (method, path, handler, store) {
         isRegex = isRegex || path[i] === '('
         if (isRegex) {
           i = path.indexOf(')', i) + 1
-          break
+          if (i) break
+
+          throw new TypeError("Invalid parameter regexp expression")
         } else if (path.charCodeAt(i) !== 45) {
           i++
         } else {
--- a/index.js
+++ b/index.js
@@ -90,7 +90,7 @@ Router.prototype._on = function _on (method, path, handler, store) {
           i = path.indexOf(')', i) + 1
           if (i) break
 
-          throw new TypeError("Invalid parameter regexp expression")
+          throw new TypeError('Invalid parameter regexp expression')
         } else if (path.charCodeAt(i) !== 45) {
           i++
         } else {
--- a/index.js
+++ b/index.js
@@ -516,6 +516,12 @@ function getClosingParenthensePosition (path, idx) {
   while (idx < path.length) {
     idx++
 
+    // ignore skipped chars
+    if (path[idx] === '\\') {
+      idx++
+      continue
+    }
+
     if (path[idx] === ')') {
       parentheses--
     } else if (path[idx] === '(') {
--- a/index.js
+++ b/index.js
@@ -87,10 +87,8 @@ Router.prototype._on = function _on (method, path, handler, store) {
       while (i < len && path.charCodeAt(i) !== 47) {
         isRegex = isRegex || path[i] === '('
         if (isRegex) {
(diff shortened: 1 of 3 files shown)


absolux:
@delvedor I'm not using my laptop, so I push directly on Github.
I have updated the test case, and now working on the fix.

absolux:
@delvedor What do you think about throwing errors for invalid paths ?

mcollina:
> What do you think about throwing errors for invalid paths?

During insertion or at runtime?

mcollina:
good for me.

index.js
```diff
@@ -88,7 +88,9 @@ Router.prototype._on = function _on (method, path, handler, store) {
         isRegex = isRegex || path[i] === '('
         if (isRegex) {
           i = path.indexOf(')', i) + 1
-          break
+          if (i) break
+
+          throw new TypeError('Invalid parameter regexp expression')
```
delvedor: Can you also add the path to the error message?
It will be easier to fix.

absolux: Done

test/issue-63.test.js
```diff
@@ -0,0 +1,14 @@
...
+const t = require('tap')
+const factory = require('../')
+
+t.test('issue-63', (t) => {
+  t.plan(1)
+
+  const fmw = factory()
+
+  t.throws(function () {
+    fmw.on('GET', '/foo/:id(a', function () {})
```
delvedor: Can you also check the expected error message?
Something like
```js
try  {
  fmw.on('GET', '/foo/:id(a', function () {})
  t.fail('should fail')
} catch (err) {
  t.is(err.message, 'Invalid regexp expression in "/foo/:id(a"')
}
```

delvedor:
@mafintosh does this fix both your issues?

index.js
```diff
@@ -496,3 +496,34 @@ function getWildcardNode (node, method, path, len) {
...
+function getClosingParenthensePosition (path, idx) {
+  // `path.indexOf()` will always return the first position of the closing parenthese,
+  // but it's inefficient for grouped or wrong regexp expressions.
+  // see issues #62 and #63 for more info
+
+  var parentheses = 1
+
+  while (idx < path.length) {
+    idx++
+
```
mafintosh: if the next char is `\` then you need to ignore the next one for it to be spec compliant :)

```
if (path[idx] === '\\') {
  continue
}
```

absolux: Hi @mafintosh can give a use case using backslash. thanks

mafintosh: Sure, `find.get(/foo/(bar/\(baz\)))`, matches '/foo/bar(baz)'.

mafintosh:
@delvedor ya and added a small comment

mafintosh:
LTGM from otherwise

absolux:
@delvedor it was a mistake, I reverted it
@mafintosh could you review the new changes

delvedor:
Landed in `v1.10.4` :)

PR #70, opened 2018-03-12
Fix route matching issue with common prefix

### LINKED ISSUE #67: URL matches unexpectedly when multiple routes have a similar prefix
#65 solved #59/#60, but it created this new bug:

```js
var router = require('find-my-way')()

router.on('GET', '/b/', function b () {})
router.on('GET', '/b/bulk', function b_bulk () {})
router.on('GET', '/b/ulk', function b_ulk () {})

console.log(router.find('GET', '/bulk')) // { handler: [Function: b_ulk], ... }
```

#65 may not have been the right fix. It seems like something is going on with how prefixes are matched and might have something to do with the `label` property on nodes.

nwoltman:
The `Node#find()` method was trying to take a shortcut by matching against the node's "label" (the first character of the prefix). This fixes route matching issues by matching against the entire prefix when the node is a static node.

Fixes #67
This is the correct fix for #59/#60

<details>
<summary>Benchmark</summary>

**before**
```
lookup static route x 29,941,458 ops/sec ±2.24% (90 runs sampled)
lookup dynamic route x 1,707,649 ops/sec ±1.97% (94 runs sampled)
lookup dynamic multi-parametric route x 974,717 ops/sec ±0.43% (94 runs sampled)
lookup dynamic multi-parametric route with regex x 889,377 ops/sec ±0.52% (96 runs sampled)
lookup long static route x 2,713,562 ops/sec ±0.42% (94 runs sampled)
find static route x 35,555,376 ops/sec ±0.81% (96 runs sampled)
find dynamic route x 2,027,441 ops/sec ±0.51% (93 runs sampled)
find dynamic multi-parametric route x 1,153,594 ops/sec ±0.32% (98 runs sampled)
find dynamic multi-parametric route with regex x 1,000,319 ops/sec ±0.51% (96 runs sampled)
find long static route x 5,245,622 ops/sec ±0.36% (95 runs sampled)
```

**after**
```
lookup static route x 32,129,986 ops/sec ±0.80% (88 runs sampled)
lookup dynamic route x 1,672,248 ops/sec ±0.45% (97 runs sampled)
lookup dynamic multi-parametric route x 930,559 ops/sec ±0.40% (92 runs sampled)
lookup dynamic multi-parametric route with regex x 840,688 ops/sec ±0.43% (97 runs sampled)
lookup long static route x 2,019,996 ops/sec ±0.40% (98 runs sampled)
find static route x 36,189,618 ops/sec ±0.71% (94 runs sampled)
find dynamic route x 2,049,103 ops/sec ±0.51% (93 runs sampled)
find dynamic multi-parametric route x 1,088,319 ops/sec ±0.64% (95 runs sampled)
find dynamic multi-parametric route with regex x 946,871 ops/sec ±0.60% (92 runs sampled)
find long static route x 3,108,926 ops/sec ±0.39% (98 runs sampled)
```

</details>

Benchmarks drop the longer the route is. This may be because now there's some duplicated prefix matching somewhere.

### DIFF
--- a/node.js
+++ b/node.js
@@ -67,7 +67,7 @@ Node.prototype.find = function (path, method) {
     var child = this.children[i]
     if (
       (child.numberOfChildren !== 0 || child.handlers[method] !== null) &&
-      (child.kind !== 0 || path.startsWith(child.prefix))
+      (child.kind !== 0 || path.slice(0, child.prefix.length) === child.prefix)
     ) {
       return child
     }
--- a/node.js
+++ b/node.js
@@ -62,14 +62,14 @@ Node.prototype.findByLabel = function (label) {
   return null
 }
 
-Node.prototype.find = function (label, method) {
+Node.prototype.find = function (path, method) {
   for (var i = 0; i < this.numberOfChildren; i++) {
     var child = this.children[i]
-    if (child.numberOfChildren !== 0 || child.handlers[method] !== null) {
-      if (child.label === label && child.kind === 0) {
-        return child
-      }
-      if (child.kind !== 0) return child
+    if (
+      (child.numberOfChildren !== 0 || child.handlers[method] !== null) &&
+      (child.kind !== 0 || path.startsWith(child.prefix))
+    ) {
+      return child
     }
   }
   return null
--- a/index.js
+++ b/index.js
@@ -275,7 +275,6 @@ Router.prototype.find = function find (method, path) {
   var pindex = 0
   var params = []
   var i = 0
-  var isStatic = true
 
   while (true) {
     var pathLen = path.length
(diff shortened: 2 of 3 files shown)


node.js
```diff
@@ -62,14 +62,14 @@ Node.prototype.findByLabel = function (label) {
...
   for (var i = 0; i < this.numberOfChildren; i++) {
     var child = this.children[i]
-    if (child.numberOfChildren !== 0 || child.handlers[method] !== null) {
-      if (child.label === label && child.kind === 0) {
-        return child
-      }
-      if (child.kind !== 0) return child
+    if (
+      (child.numberOfChildren !== 0 || child.handlers[method] !== null) &&
+      (child.kind !== 0 || path.startsWith(child.prefix))
```
delvedor: I think `startsWith` is causing the performances drop.
Can you try with `path.slice(0, child.prefix.length) === child.prefix`?

mcollina:
If we want to speed up string checking, we can also use a XOR as done in here https://github.com/mcollina/bloomrun/blob/master/lib/safeEqual.js.

delvedor:
A benchmark comparison could be nice!

nwoltman:
@delvedor Your suggestion did improve the performance a bit. I updated the benchmarks.

@mcollina I tried implementing a custom `startsWith` function that looked similar to that but with `===` and it was slower than the native `string.startsWith()`. Looping across char codes got slowed down in TurboFan, but it should be fast again with V8 66 (reference: https://bugs.chromium.org/p/v8/issues/detail?id=7326&desc=2).

PR #71, opened 2018-03-13
Refactored Node object

delvedor:
As titled.

The benchmarks are almost the same except for the long static route, with this refactor we gain around 400-500k ops/sec more.
We must figure out a way to optimize the dynamic route handling.

cc @nwoltman

### DIFF
--- a/bench.js
+++ b/bench.js
@@ -8,6 +8,7 @@ const FindMyWay = require('./')
 const findMyWay = new FindMyWay()
 findMyWay.on('GET', '/', () => true)
 findMyWay.on('GET', '/user/:id', () => true)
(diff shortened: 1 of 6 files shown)


delvedor:
I've replaced `decodeURIComponent` with [fast-decode-uri-component](https://github.com/delvedor/fast-decode-uri-component), which is way faster and let us gain around 80% ops/sec more on dynamic routes.

mcollina:
Can you update the benchmarks?

delvedor:
@mcollina I'll update the [benchmarks](https://github.com/delvedor/router-benchmark) repo.

PR #72, opened 2018-03-13
Use http methods

allevo:
As titled

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,5 @@ jspm_packages
 
 # vim swap files
 *.swp
+
+package-lock.json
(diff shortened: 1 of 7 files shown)


delvedor:
You should also update this code:
https://github.com/delvedor/find-my-way/blob/cc6ac4e5a1b1617ca2d84cce33201f727ae12c8b/node.js#L129-L140

Same for the shorthand api.
And maybe add some test.

allevo:
I haven't added shorthand api because some methods conflict with the existent API like `find`.

delvedor:
> Let me know if I have to add them

Yes please, an half complete API is not that good, even if is a simple thing :)

allevo:
done. Changed README too

PR #74, opened 2018-03-21
Fix lookups of unsupported methods

nwoltman:
Currently, looking up an unsupported is throwing an error because the method isn't in the `handlers` object, so trying to get a handler from the `handlers` object will return `undefined` (and currently only `null` is being handled).

### DIFF
--- a/index.js
+++ b/index.js
@@ -290,7 +290,7 @@ Router.prototype.find = function find (method, path) {
     // found the route
     if (pathLen === 0 || path === prefix) {
       var handle = currentNode.handlers[method]
-      if (handle !== null) {
+      if (handle !== null && handle !== undefined) {
         var paramsObj = {}
         if (handle.paramsLength > 0) {
           var paramNames = handle.params
@@ -455,7 +455,7 @@ function getWildcardNode (node, method, path, len) {
   var decoded = fastDecode(path.slice(-len))
   if (decoded === null) return null
   var handle = node.handlers[method]
-  if (handle !== null) {
+  if (handle !== null && handle !== undefined) {
     return {
       handler: handle.handler,
       params: { '*': decoded },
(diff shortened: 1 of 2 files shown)


delvedor:
@nwoltman we are using `http.METHODS` from Node core, what happens if you send to a basic node http server a request with the `TROLL` method?

nwoltman:
With the `http` and `https` modules, Node responds with a `400 Bad Request`. But with the `http2` module, the request goes through just like any other (if the connection uses HTTP/2).

delvedor:
> With the http and https modules, Node responds with a 400 Bad Request. 

So `find-my-way` is never executed?

> But with the http2 module, the request goes through just like any other (if the connection uses HTTP/2).

I'm not sure if this correct, maybe there is some issue in core. @mcollina?

nwoltman:
Yes, with HTTP/1.x `find-my-way` is never executed, but HTTP/2 seems to allow any method name.

jasnell:
This is a variance in the way `http2` handles http methods vs. the `http` and `https` modules but it's not a bug.

The HTTP specification defines methods as being extensible. Previously, Node.js has limited the actual HTTP methods that may be used to a fixed set largely because the parsing logic used by the http_parser is difficult to optimize to support any arbitrary method name. With the `http2` implementation, however, this restriction disappears and we are now able to support any method. What this means in code, however, is that application logic must pay attention to the method that they are actually receiving from clients and handle that appropriately on their own. It would be relatively trivial to use a middleware for this in most frameworks.

FWIW, `TROLL`is a valid HTTP method per the syntax... it's just not a *registered* method with any standard semantics. So, `TROLL /whatever HTTP/1.1` is a *valid* HTTP request (in either HTTP/1 and HTTP/2.

delvedor:
@jasnell thanks!

PR #79, opened 2018-06-08
Added support for semver routes

delvedor:
As titled 🎉 

There is no impact on the actual performances 🚀 
But if you use a `semver` route you will encounter a huge drop in the performances, from 25 million to 110 thousand ops/sec on my machine. This is due to the `semver` module which is not very performant, we could think about writing one of our own in the future.

Feedbacks?

cc @hekike @jsumners @nwoltman


test/issue-28.test.js
```diff
@@ -359,7 +359,7 @@ test('Nested wildcards with parametric and static - 2', t => {
   })
 
   findMyWay.lookup(
-    { method: 'GET', url: '/foo3/bar1' },
+    { method: 'GET', url: '/foo3/bar1', headers: {} },
```
mcollina: is this change `headers: {}`  needed in all those cases?

delvedor: Yes, because `lookup` will directly search in `req.headers['accept-version']`.

delvedor:
Since the performances of the `semver` module are low, I've built [`semver-store`](https://github.com/delvedor/semver-store), which give us an huge gain when using semver based routes, from 110k ops/sec to 4.5 millions.

`semver-store` supports only basic semver functionalities, so *advanced ranges* and *pre-releases* currently are not supported, while basic ranges such as `1.x` or `2.3.x` are supported with great performances!

test/issue-28.test.js
delvedor: Nope, `lookup` accepts node *request* and *response* objects, in the test we are using a basic object for simplicity.

delvedor:
As an option in the route configuration I guess.
I'll open a pr :)

README.md
```diff
@@ -90,7 +90,26 @@ router.on('GET', '/example', (req, res, params, store) => {
...
+  res.send('Hello from 1.2.0!')
+})
+
+router.on('GET', '/example', { version: '2.4.0' }, (req, res, params) => {
+  res.send('Hello from 2.4.0!')
+})
+
+// The 'Accept-Version' header could be '1.2.0' as well as '2.x' or '2.4.x'
+```
+If you declare multiple version with the same *major* or *minor* `find-my-way` will always choose the highest compatible with the `Accept-Version` header value.
```
nwoltman: version -> versions

hekike:
Awesome, thanks for implementing it! 🎉

jsumners:
I think it's far easier to just namespace things via paths. This also goes against the opening selling point for `find-my-way`: "A crazy fast HTTP router...".

delvedor:
> I think it's far easier to just namespace things via paths. 

@jsumners you can't namespace the routes, because that solution does not support *semver ranges*, which is all this pr is about.

> This also goes against the opening selling point for find-my-way: "A crazy fast HTTP router...".

That's not true, the standard use will continue to work at speed as always.
Regarding the versioned routes obviously you must pay a cost, but thanks to [semver-store](https://github.com/delvedor/semver-store) is not as bad as it was with [semver](https://github.com/npm/node-semver).

PR #81, opened 2018-06-26
prevent looping over prototype chain

jackhftang:
Some libraries add function to (Iteratable) Array.prototype and for(var .. in ..) loop will iterate over those function and code will throw syntax error if the function clousure has more than one line.


mcollina:
Thanks for catching this! 

Linting in failing on CI, would you mind fixing it?

```
/home/travis/build/delvedor/find-my-way/node.js:217:13: Infix operators must be spaced.
```

PR #84, opened 2018-06-30
Add an option to disable case-sensitive routing

mcollina:
See https://github.com/fastify/fastify/issues/1007.

@thearegee would this work for you?

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,6 +32,7 @@ function Router (opts) {
     this.defaultRoute = null
   }
 
+  this.caseSensitive = opts.caseSensitive === undefined ? true : opts.caseSensitive
   this.ignoreTrailingSlash = opts.ignoreTrailingSlash || false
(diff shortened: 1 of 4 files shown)


thearegee:
@mcollina this looks perfect, could just do with a note maybe in the README? 
Just so people know its available

mcollina:
Done.

index.js
```diff
@@ -308,6 +320,9 @@ Router.prototype.lookup = function lookup (req, res) {
 }
 
 Router.prototype.find = function find (method, path, version) {
+  if (!this.caseSensitive) {
```
delvedor: `if (this.caseSensitive === false)?`
It should make V8 happier (but I'm not super sure because the value is already a Boolean).

mcollina: updated.

mcollina:
@delvedor updated.

PR #85, opened 2018-07-02
Update semver

delvedor:
As titled :D

cc @AVVS


AVVS:
Awesome :)

delvedor:
From a code perspective support that will increase the code complexity, and I don't know if it is worth.
It's a different method of routing, allowing this mixed logic could lead to unexpected bugs IMHO.

PR #89, opened 2018-09-06
Use generated radix tree

mafintosh:
This makes the router use a code generated radix tree instead.
Main things missing:

* [ ] Add missing options (such as case insensitive etc, shouldn't be hard at all)
* [ ] Fix regex issue. I'm not sure how your regex matcher works. I just do an eager match against the pattern passed, but you support `$` as an inplace regex which I'm unsure what means
* [ ] Remove dead code. We can prob remove 75% of the code :D
* [ ] Versioned routes. Unsure how these work.
* [ ] Don't compile on each .on, but do it nextTick or something?

Benchmarks

## Before

```
lookup static route x 56,675,717 ops/sec ±3.91% (92 runs sampled)
lookup dynamic route x 3,907,986 ops/sec ±1.61% (92 runs sampled)
lookup dynamic multi-parametric route x 2,094,830 ops/sec ±1.50% (94 runs sampled)
lookup dynamic multi-parametric route with regex x 1,559,524 ops/sec ±0.91% (97 runs sampled)
lookup long static route x 3,840,774 ops/sec ±0.72% (94 runs sampled)
lookup long dynamic route x 2,517,142 ops/sec ±0.56% (95 runs sampled)
find static route x 46,681,402 ops/sec ±0.79% (97 runs sampled)
find dynamic route x 4,471,999 ops/sec ±1.18% (90 runs sampled)
find dynamic multi-parametric route x 2,275,319 ops/sec ±1.89% (91 runs sampled)
find dynamic multi-parametric route with regex x 1,636,727 ops/sec ±1.38% (92 runs sampled)
find long static route x 5,429,251 ops/sec ±0.88% (92 runs sampled)
find long dynamic route x 3,462,368 ops/sec ±0.73% (90 runs sampled)
```

## After

```
lookup static route x 52,844,596 ops/sec ±0.64% (91 runs sampled)
lookup dynamic route x 9,299,696 ops/sec ±1.13% (91 runs sampled)
lookup dynamic multi-parametric route x 5,123,525 ops/sec ±1.20% (92 runs sampled)
lookup dynamic multi-parametric route with regex x 2,952,715 ops/sec ±0.91% (92 runs sampled)
lookup long static route x 7,437,460 ops/sec ±1.05% (95 runs sampled)
lookup long dynamic route x 5,207,066 ops/sec ±0.87% (97 runs sampled)
find static route x 72,317,353 ops/sec ±0.88% (96 runs sampled)
find dynamic route x 14,775,400 ops/sec ±1.96% (91 runs sampled)
find dynamic multi-parametric route x 6,813,272 ops/sec ±0.91% (92 runs sampled)
find dynamic multi-parametric route with regex x 3,257,229 ops/sec ±1.11% (93 runs sampled)
find long static route x 14,222,061 ops/sec ±3.74% (88 runs sampled)
find long dynamic route x 10,972,760 ops/sec ±0.62% (93 runs sampled)
```

(Missing versioned routes as it's not impl yet, but most of these are >2x faster :))

### DIFF
--- a/bench.js
+++ b/bench.js
@@ -12,7 +12,7 @@ findMyWay.on('GET', '/user/:id/static', () => true)
 findMyWay.on('GET', '/customer/:name-:surname', () => true)
 findMyWay.on('GET', '/at/:hour(^\\d+)h:minute(^\\d+)m', () => true)
 findMyWay.on('GET', '/abc/def/ghi/lmn/opq/rst/uvz', () => true)
-findMyWay.on('GET', '/', { version: '1.2.0' }, () => true)
+// findMyWay.on('GET', '/', { version: '1.2.0' }, () => true)
 
 suite
   .add('lookup static route', function () {
@@ -33,9 +33,9 @@ suite
   .add('lookup long dynamic route', function () {
     findMyWay.lookup({ method: 'GET', url: '/user/qwertyuiopasdfghjklzxcvbnm/static', headers: {} }, null)
   })
-  .add('lookup static versioned route', function () {
-    findMyWay.lookup({ method: 'GET', url: '/', headers: { 'accept-version': '1.x' } }, null)
-  })
+//  .add('lookup static versioned route', function () {
+//    findMyWay.lookup({ method: 'GET', url: '/', headers: { 'accept-version': '1.x' } }, null)
+//  })
   .add('find static route', function () {
     findMyWay.find('GET', '/', undefined)
   })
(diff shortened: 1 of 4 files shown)


mcollina:
>  Fix regex issue. I'm not sure how your regex matcher works. I just do an eager match against the pattern passed, but you support $ as an inplace regex which I'm unsure what means

These are express styles regexps, with parameters. such as:

```
Route path: /flights/:from-:to
Request URL: http://localhost:3000/flights/LAX-SFO
req.params: { "from": "LAX", "to": "SFO" }
```

But also

```
Route path: /user/:userId(\d+)
Request URL: http://localhost:3000/user/42
req.params: {"userId": "42"}
```

Turn out that it is one of the most depended-on express features.

---

cc @allevo you played with this, you might want to help out1

allevo:
> cc @allevo you played with this, you might want to help out1

I'd like to be focused on [the router I'm writting](https://www.npmjs.com/package/road-marking) that seems to be a lot faster that `find-my-way`.

Anyway all supported feature should be kept even if the performance decreases.

mafintosh:
@mcollina does the regex only match until the next '/'? That was the part I was unsure about

mcollina:
yes, it does.
Il giorno gio 6 set 2018 alle 13:15 Mathias Buus <notifications@github.com>
ha scritto:

mcollina:
Thanks! Closing for no activity

PR #91, opened 2018-09-07
Prevent TypeError: m.toLowerCase is not a function

TheNoim:
Never forget this, if you use for..in


TheNoim:
this escalated quickly

TheNoim:
Added a test @mcollina

PR #92, opened 2018-09-07
Update readme for explaining better the route order

allevo:
as titled

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,9 @@ Having a route with multiple parameters may affect negatively the performance, s
 
 <a name="match-order"></a>
 ##### Match order
-The routes are matched in the following order:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -168,7 +168,9 @@ Having a route with multiple parameters may affect negatively the performance, s
 
 <a name="match-order"></a>
 ##### Match order
-The routes are matched in the following order:
+The routing algorithm tokenizes the request path into segments.
```
mcollina: (divided by /).

allevo: => `The routing algorithm tokenizes (use the '/' to split) the request path into segments.`??

mcollina: not perfect but it’ll work.

allevo:
Update as suggested

PR #100, opened 2018-10-13
Bumped dependencies

davisjam:
Upgrade to the hot-off-the-press safe-regex v2.0.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ node_js:
   - "10"
   - "8"
   - "6"
-  - "4"
 
 after_script:
   - npm run coveralls
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
   },
   "dependencies": {
     "fast-decode-uri-component": "^1.0.0",
-    "safe-regex": "^1.1.0",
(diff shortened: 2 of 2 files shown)


davisjam:
Sorry, haven't yet published this package due to weird npm misconfiguration on my end.

Will update the PR soon.

davisjam:
PR updated. Let's see how Travis-CI does.

mcollina:
what’s the main difference with the new version?

davisjam:
1. Switches from `ret` to `regexp-tree` to build a regex AST. I didn't notice any differences in the test suite (or in the test suites of a few dependencies) but it is possible that some regexes may no longer be parse-able and some others may be newly parse-able.
2. Fixes a class of false negatives for which star height was incorrectly calculated. For example regexes like `(a*|b)*` -- see [substack's safe-regex #8](https://github.com/substack/safe-regex/issues/8).

mcollina:
CI fails on Node v4. Can you just remove it from travis.yml? This is semver/major anyway.

davisjam:
Done.

delvedor:
Done in v2, thanks for letting us know!

PR #102, opened 2018-12-05
chore: add typescript typings

disintegrator:
Hi there, I'm considering using this library in my project and thought I'd contribute my typings for it. These typings should work for TypeScript version 3.x.x

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -12,13 +12,40 @@ declare namespace Router {
   }
 
   type HTTPMethod =
+    | 'ACL'
+    | 'BIND'
+    | 'CHECKOUT'
+    | 'CONNECT'
+    | 'COPY'
     | 'DELETE'
     | 'GET'
     | 'HEAD'
+    | 'LINK'
+    | 'LOCK'
+    | 'M-SEARCH'
+    | 'MERGE'
+    | 'MKACTIVITY'
+    | 'MKCALENDAR'
+    | 'MKCOL'
+    | 'MOVE'
+    | 'NOTIFY'
+    | 'OPTIONS'
     | 'PATCH'
     | 'POST'
+    | 'PROPFIND'
+    | 'PROPPATCH'
+    | 'PURGE'
     | 'PUT'
-    | 'OPTIONS';
+    | 'REBIND'
+    | 'REPORT'
+    | 'SEARCH'
+    | 'SOURCE'
+    | 'SUBSCRIBE'
+    | 'TRACE'
+    | 'UNBIND'
+    | 'UNLINK'
+    | 'UNLOCK'
+    | 'UNSUBSCRIBE';
 
   type Handler<V extends HTTPVersion> = (
     req: V extends HTTPVersion.V1 ? IncomingMessage : Http2ServerRequest,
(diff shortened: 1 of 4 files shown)


index.d.ts
```diff
@@ -0,0 +1,116 @@
...
+  config?: Router.Config<V>
+): Router.Instance<V>;
+
+declare namespace Router {
+  enum HTTPVersion {
+    V1 = 'http1',
+    V2 = 'http2'
+  }
+
+  type HTTPMethod =
```
delvedor: We support all HTTP methods, as you can see [here](https://github.com/delvedor/find-my-way/blob/90348f3e4b92e1c52bc6be77b1e17b68eb5b5040/index.js#L488-L498).
This list probably needs to be updated.

index.d.ts
```diff
@@ -0,0 +1,116 @@
...
+    | 'PATCH'
+    | 'POST'
+    | 'PUT'
+    | 'OPTIONS';
+
+  type Handler<V extends HTTPVersion> = (
+    req: V extends HTTPVersion.V1 ? IncomingMessage : Http2ServerRequest,
+    res: V extends HTTPVersion.V1 ? ServerResponse : Http2ServerResponse,
+    params: { [k: string]: string | undefined },
+    store: any
```
delvedor: I guess `params` and `store` can be optional.

disintegrator: I noticed when using the library in a node repl that you always return params. It's just an empty object if there are no params in matched route. As for store, since its opaque, I had to use any which trumps optional/nullable (`any | null === any`, `any | undefined === any`, `store?: any === store: any`)

disintegrator: Note the value of params for different cases in the screenshot below:
image

delvedor: Yes, I'm always returning them, but what I mean is that you don't have to declare them always, with this declaration you have to, right?
```js
router.on('GET', '/', (req, res) => {}) // this is perfectly fine
```

disintegrator:
@delvedor tests are now up for review!

index.d.ts
delvedor: Good to know, thanks!

index.d.ts
```diff
@@ -0,0 +1,170 @@
...
+    subscribe: ShortHandRoute<V>;
+    trace: ShortHandRoute<V>;
+    unbind: ShortHandRoute<V>;
+    unlink: ShortHandRoute<V>;
+    unlock: ShortHandRoute<V>;
+    unsubscribe: ShortHandRoute<V>;
+  }
+}
+
+export = Router;
```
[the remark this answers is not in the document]
disintegrator: I think changing to export default will make this fail in commonjs codebases. Im on my phone so can’t confirm this yet but I recall seeing this problem when typing another library that used `module.export`.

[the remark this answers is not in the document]
disintegrator: The way I have it declared will allow the library to be consumed in es and commonjs environments:
```
const Router = require('find-my-way') // should work fine

import * as Router from 'find-my-way' // should work fine

// with esModuleInterop
import Router from 'find-my-way' // should work fine
```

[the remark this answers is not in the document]
disintegrator: The follow is a preview of what happens if we switch to `export default`

image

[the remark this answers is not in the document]
delvedor: Great, thanks!

delvedor:
Landed in [`v1.16.0`](https://github.com/delvedor/find-my-way/releases/tag/v1.16.0), sorry for the delay! :)

PR #103, opened 2018-12-08
feat: allow setting handler scope using lookup

disintegrator:
I'd like to suggest allowing the lookup method to accept a third state/context parameter that becomes the bound scope of a handler.

**Proposed change:**
`lookup(request, response, [context])`

```js
router.on('GET', '*', function(req, res) {
  res.end(this.greeting);
})
router.lookup(req, res, { greeting: 'Hello, World!' })
```

**Background:**
I'm using a couple of find-my-way routers in a larger fastify application and need a way to invoke `router.lookup` on these nested routers while also passing some extra context derived in my application from incoming requests.


disintegrator:
@mcollina sure thing. Results I have:
```
lookup static route x 25,069,326 ops/sec ±0.94% (88 runs sampled)
lookup dynamic route x 3,351,305 ops/sec ±1.38% (86 runs sampled)
lookup dynamic multi-parametric route x 1,971,455 ops/sec ±1.93% (86 runs sampled)
lookup dynamic multi-parametric route with regex x 1,519,641 ops/sec ±0.35% (95 runs sampled)
lookup long static route x 3,586,145 ops/sec ±1.00% (91 runs sampled)
lookup long dynamic route x 2,428,218 ops/sec ±1.05% (95 runs sampled)
lookup static versioned route x 9,381,271 ops/sec ±0.85% (94 runs sampled)
find static route x 42,282,253 ops/sec ±0.77% (94 runs sampled)
find dynamic route x 4,502,530 ops/sec ±0.81% (90 runs sampled)
find dynamic multi-parametric route x 2,353,420 ops/sec ±0.75% (95 runs sampled)
find dynamic multi-parametric route with regex x 1,642,743 ops/sec ±1.58% (92 runs sampled)
find long static route x 5,025,655 ops/sec ±0.71% (94 runs sampled)
find long dynamic route x 3,363,116 ops/sec ±1.30% (89 runs sampled)
find static versioned route x 12,211,408 ops/sec ±0.82% (91 runs sampled)
```

System info:
MacBook Pro (15-inch, 2017)
2.9 GHz Intel Core i7
16 GB 2133 MHz LPDDR3
Node.js v10.14.1

mcollina:
Can you also run master for comparison?

disintegrator:
Ah sorry I forgot about that...
Yikes, it looks like a cause a major dip in the "lookup static route" case
```diff
-delvedor:master
+disintegrator:lookup-with-optional-scope

-lookup static route x 42,385,194 ops/sec ±1.05% (85 runs sampled)
+lookup static route x 25,678,947 ops/sec ±0.59% (93 runs sampled)
-lookup dynamic route x 3,161,707 ops/sec ±1.42% (86 runs sampled)
+lookup dynamic route x 3,506,276 ops/sec ±1.33% (91 runs sampled)
-lookup dynamic multi-parametric route x 1,850,342 ops/sec ±1.32% (88 runs sampled)
+lookup dynamic multi-parametric route x 1,946,542 ops/sec ±1.27% (86 runs sampled)
-lookup dynamic multi-parametric route with regex x 1,421,410 ops/sec ±0.41% (92 runs sampled)
+lookup dynamic multi-parametric route with regex x 1,470,873 ops/sec ±0.41% (94 runs sampled)
-lookup long static route x 3,125,178 ops/sec ±1.70% (89 runs sampled)
+lookup long static route x 3,308,128 ops/sec ±0.68% (91 runs sampled)
-lookup long dynamic route x 2,244,370 ops/sec ±0.98% (90 runs sampled)
+lookup long dynamic route x 2,362,879 ops/sec ±1.00% (91 runs sampled)
-lookup static versioned route x 8,249,073 ops/sec ±0.62% (92 runs sampled)
+lookup static versioned route x 9,191,771 ops/sec ±0.49% (92 runs sampled)
-find static route x 37,492,501 ops/sec ±0.96% (89 runs sampled)
+find static route x 37,180,927 ops/sec ±6.95% (82 runs sampled)
-find dynamic route x 4,051,598 ops/sec ±0.84% (89 runs sampled)
+find dynamic route x 3,852,229 ops/sec ±2.52% (86 runs sampled)
-find dynamic multi-parametric route x 2,121,916 ops/sec ±0.89% (90 runs sampled)
+find dynamic multi-parametric route x 2,169,089 ops/sec ±0.76% (90 runs sampled)
-find dynamic multi-parametric route with regex x 1,515,172 ops/sec ±1.54% (92 runs sampled)
+find dynamic multi-parametric route with regex x 1,485,392 ops/sec ±1.86% (89 runs sampled)
-find long static route x 4,603,272 ops/sec ±0.56% (93 runs sampled)
+find long static route x 4,685,861 ops/sec ±0.83% (90 runs sampled)
-find long dynamic route x 2,971,945 ops/sec ±1.41% (90 runs sampled)
+find long dynamic route x 3,050,995 ops/sec ±1.28% (89 runs sampled)
-find static versioned route x 10,495,483 ops/sec ±0.70% (91 runs sampled)
+find static versioned route x 10,876,192 ops/sec ±2.74% (87 runs sampled)
```

disintegrator:
Ah sorry I forgot about that...
Yikes, it looks like a cause a major dip in the "lookup static route" case
```diff
-delvedor:master
+disintegrator:lookup-with-optional-scope

-lookup static route x 42,385,194 ops/sec ±1.05% (85 runs sampled)
+lookup static route x 25,678,947 ops/sec ±0.59% (93 runs sampled)
-lookup dynamic route x 3,161,707 ops/sec ±1.42% (86 runs sampled)
+lookup dynamic route x 3,506,276 ops/sec ±1.33% (91 runs sampled)
-lookup dynamic multi-parametric route x 1,850,342 ops/sec ±1.32% (88 runs sampled)
+lookup dynamic multi-parametric route x 1,946,542 ops/sec ±1.27% (86 runs sampled)
-lookup dynamic multi-parametric route with regex x 1,421,410 ops/sec ±0.41% (92 runs sampled)
+lookup dynamic multi-parametric route with regex x 1,470,873 ops/sec ±0.41% (94 runs sampled)
-lookup long static route x 3,125,178 ops/sec ±1.70% (89 runs sampled)
+lookup long static route x 3,308,128 ops/sec ±0.68% (91 runs sampled)
-lookup long dynamic route x 2,244,370 ops/sec ±0.98% (90 runs sampled)
+lookup long dynamic route x 2,362,879 ops/sec ±1.00% (91 runs sampled)
-lookup static versioned route x 8,249,073 ops/sec ±0.62% (92 runs sampled)
+lookup static versioned route x 9,191,771 ops/sec ±0.49% (92 runs sampled)
-find static route x 37,492,501 ops/sec ±0.96% (89 runs sampled)
+find static route x 37,180,927 ops/sec ±6.95% (82 runs sampled)
-find dynamic route x 4,051,598 ops/sec ±0.84% (89 runs sampled)
+find dynamic route x 3,852,229 ops/sec ±2.52% (86 runs sampled)
-find dynamic multi-parametric route x 2,121,916 ops/sec ±0.89% (90 runs sampled)
+find dynamic multi-parametric route x 2,169,089 ops/sec ±0.76% (90 runs sampled)
-find dynamic multi-parametric route with regex x 1,515,172 ops/sec ±1.54% (92 runs sampled)
+find dynamic multi-parametric route with regex x 1,485,392 ops/sec ±1.86% (89 runs sampled)
-find long static route x 4,603,272 ops/sec ±0.56% (93 runs sampled)
+find long static route x 4,685,861 ops/sec ±0.83% (90 runs sampled)
-find long dynamic route x 2,971,945 ops/sec ±1.41% (90 runs sampled)
+find long dynamic route x 3,050,995 ops/sec ±1.28% (89 runs sampled)
-find static versioned route x 10,495,483 ops/sec ±0.70% (91 runs sampled)
+find static versioned route x 10,876,192 ops/sec ±2.74% (87 runs sampled)
```

Do you think it's a good idea to conditionally set the handler scope if ctx is passed? So like this:
```
  return ctx == null
    ? handle.handler(req, res, handle.params, handle.store)
    : handle.handler.call(ctx, req, res, handle.params, handle.store)
```

This result in the following
```diff
-delvedor:master
+candidate

-lookup static route x 48,311,923 ops/sec ±1.26% (91 runs sampled)
+lookup static route x 49,225,516 ops/sec ±0.66% (94 runs sampled)

**Comment by mcollina** [comment]
I'd be ok with that change.

**Comment by disintegrator** [comment]
@mcollina / @delvedor, thanks for the feedback. I've updated the implementation and TS typings. The PR description now contains the latest benchmark comparisons with master.

File: index.js [hunk 1]
```diff
@@ -317,10 +317,12 @@ Router.prototype.off = function off (method, path) {
   })
 }
 
-Router.prototype.lookup = function lookup (req, res) {
+Router.prototype.lookup = function lookup (req, res, ctx) {
   var handle = this.find(req.method, sanitizeUrl(req.url), req.headers['accept-version'])
-  if (handle === null) return this._defaultRoute(req, res)
-  return handle.handler(req, res, handle.params, handle.store)
+  if (handle === null) return this._defaultRoute(req, res, ctx)
+  return ctx == null
```
delvedor: I think here we can use a stricter check :)
```suggestion
  return ctx === undefined
```

index.js
```diff
@@ -472,9 +474,11 @@ Router.prototype.find = function find (method, path, version) {
   }
 }
 
-Router.prototype._defaultRoute = function (req, res) {
+Router.prototype._defaultRoute = function (req, res, ctx) {
   if (this.defaultRoute !== null) {
-    this.defaultRoute(req, res)
+    return ctx == null
```
delvedor: Same here :)
```suggestion
    return ctx === undefined
```

disintegrator: Suggestion applied!

test/lookup.test.js
```diff
@@ -0,0 +1,59 @@
...
+})
+
+test('lookup calls default route handler with context as scope', t => {
+  t.plan(1)
+
+  const ctx = { foo: 'bar' }
+
+  const findMyWay = FindMyWay({
+    defaultRoute (req, res) {
+      t.equal(this, ctx)
```
delvedor: Here you should use `deepEqual` :D
```suggestion
      t.deepEqual(this, ctx)
```

delvedor: The same in all the other occurrences, when you are comparing objects you should always use `deepEqual` :)

disintegrator: I use equal deliberately because I wanted ensure referential equality. It doesn’t seem deepEqual does the same or am I mistaken?

mcollina: It doesn’t, equal is ok.

PR #108, opened 2019-01-08
Fix fastify 1374

mcollina:
Fixes https://github.com/fastify/fastify/issues/1374.

Benchmarks stays consistent before and after.


mcollina:
The regexp is safe.

mcollina:
Added a check on require to validate it's safe.

PR #109, opened 2019-01-16
Handle request moved to separate method

cravler:


cravler:
I want to write a middleware for [koa](https://github.com/koajs/koa) and with this changes it will be much easier

```js
class KoaRouter extends Router {
        _defaultRoute(ctx, next) {
            if (this.defaultRoute) {
                return this.defaultRoute(ctx, next);
            }

            return next();
        }

        lookup() {
            const self = this;

            return function (ctx, next) {
                const handle = self.handle(ctx.req);

                if (handle === null) {
                    return self._defaultRoute(ctx, next);
                }

                ctx.params = handle.params;
                ctx.store = handle.store;

                return handle.handler(ctx, next);
            }
        }
    }
```

mcollina:
we expose `find()` for that specific reason.

cravler:
But then I do not have access to `sanitizeUrl` form my middleware and I need to copy `find-my-way` internal logic that handling request and I need do it allways when somthing will be changed, I think it is bad.

delvedor:
Then probably it is better to expose `sanitizeUrl` :)

cravler:
And how about this logic `this.find(req.method, sanitizeUrl(req.url), req.headers['accept-version'])`?

mcollina:
Some notes: `find` is a low-level API that allows a certain degree of flexibility, `lookup` is an opinionated API that removes some flexibility. You would like to add `handle`, which is an in-between logic. I'm already not happy about having both lookup and find, and I'd prefer to _remove_ `lookup` vs adding `handle`.

+1 in exposing sanitizeUrl as an utility.

cravler:
@mcollina, by removing `lookup` we break BC, better to add `deprecated` on this method and remove it with `v2`.

cravler:
If we add `handle`, usage will look like:

```js
const http = require('http');
const router = require('find-my-way')();

router.on('GET', '/', (req, res, params) => {
    res.end('{"message":"hello world"}');
});

    const result = router.handle(req);
    if (null !== result) {
        return result.handler(req, res, result.params, result.store);
    }
    res.statusCode = 404;
    res.end();
});

server.listen(3000, err => {
    if (err) {
        throw err;
    }
    console.log('Server listening on: http://localost:3000');
});
```

Looks more clear and no hardcode for handlers. No need for 'lookup' & 'defaultRoute'.

delvedor:
Hello! Closing this for now, feel free to reopen if you want to discuss it more!

PR #111, opened 2019-02-07
Falling back to node's parametricBrother

IpaliboWhyte:
This is causing issues resulting in nodes not having a parametric brother as witnessed in #110 

`node bench.js` result:

**Master**
```
lookup static route x 36,805,474 ops/sec ±1.94% (83 runs sampled)
lookup dynamic route x 3,062,199 ops/sec ±1.85% (87 runs sampled)
lookup dynamic multi-parametric route x 1,681,604 ops/sec ±1.76% (85 runs sampled)
lookup dynamic multi-parametric route with regex x 1,319,392 ops/sec ±1.05% (88 runs sampled)
lookup long static route x 2,975,863 ops/sec ±0.85% (90 runs sampled)
lookup long dynamic route x 2,116,684 ops/sec ±0.90% (91 runs sampled)
lookup static versioned route x 7,842,228 ops/sec ±0.87% (91 runs sampled)
find static route x 38,501,099 ops/sec ±0.69% (94 runs sampled)
find dynamic route x 4,113,594 ops/sec ±0.60% (93 runs sampled)
find dynamic multi-parametric route x 2,103,203 ops/sec ±0.49% (92 runs sampled)
find dynamic multi-parametric route with regex x 1,477,187 ops/sec ±1.99% (89 runs sampled)
find long static route x 4,517,911 ops/sec ±0.66% (92 runs sampled)
find long dynamic route x 3,026,659 ops/sec ±1.22% (92 runs sampled)
find static versioned route x 9,621,962 ops/sec ±1.94% (83 runs sampled)
```

**issue-110**

```
lookup static route x 37,693,101 ops/sec ±1.50% (84 runs sampled)
lookup dynamic route x 2,985,300 ops/sec ±1.75% (88 runs sampled)
lookup dynamic multi-parametric route x 1,671,388 ops/sec ±1.52% (87 runs sampled)
lookup dynamic multi-parametric route with regex x 1,278,199 ops/sec ±0.93% (88 runs sampled)
lookup long static route x 2,951,359 ops/sec ±1.04% (88 runs sampled)
lookup long dynamic route x 2,093,028 ops/sec ±1.35% (91 runs sampled)
lookup static versioned route x 7,783,319 ops/sec ±1.04% (88 runs sampled)
find static route x 35,585,628 ops/sec ±1.36% (88 runs sampled)
find dynamic route x 3,862,732 ops/sec ±1.30% (86 runs sampled)
[...]

PR #112, opened 2019-02-07
Lock tap to v12.0.1

mcollina:
Tap 12.0.1 still support Node v4. Latest versions do not anymore.


IpaliboWhyte:
I believe v12.4.1 has the latest support for Node v4

Also I added this in my PR :)

https://github.com/delvedor/find-my-way/pull/111/commits/cbe892dc3193c7086cb3d81eb7c72bcbbac46689

PR #99, opened 2018-10-04
preserve param values case in case insensitive mode

misterdjules:
The implementation is a bit crude and simplistic, and the style is not great, but I thought it would be a good start to discuss and maybe iterate on.

Also note that this PR removes leftover `console.log` statements from the previous PR, my apologies for that.

PR #115, opened 2019-02-12
preserve param values case in case insensitive mode

misterdjules:
This PR has the same goal as https://github.com/delvedor/find-my-way/pull/99, but is solving the problem using a different approach in an effort to get a better performance profile.

Using node v10.14.1 I get the following results on the current tip of the master branch:

```
$ node bench.js 
lookup static route x 47,169,425 ops/sec ±0.44% (92 runs sampled)
lookup dynamic route x 3,518,794 ops/sec ±1.27% (94 runs sampled)
lookup dynamic multi-parametric route x 1,959,970 ops/sec ±1.18% (91 runs sampled)
lookup dynamic multi-parametric route with regex x 1,485,289 ops/sec ±1.04% (95 runs sampled)
lookup long static route x 3,359,776 ops/sec ±0.51% (93 runs sampled)
lookup long dynamic route x 2,326,569 ops/sec ±0.88% (94 runs sampled)
lookup static versioned route x 8,826,650 ops/sec ±0.36% (94 runs sampled)
find static route x 40,171,616 ops/sec ±0.36% (94 runs sampled)
find dynamic route x 4,323,841 ops/sec ±0.46% (94 runs sampled)
find dynamic multi-parametric route x 2,230,310 ops/sec ±0.95% (91 runs sampled)
find dynamic multi-parametric route with regex x 1,612,491 ops/sec ±1.48% (94 runs sampled)
find long static route x 5,095,209 ops/sec ±0.49% (92 runs sampled)
find long dynamic route x 3,274,189 ops/sec ±1.00% (93 runs sampled)
find static versioned route x 12,199,768 ops/sec ±0.47% (91 runs sampled)
$
```

and the following results with the changes in this PR:

```
$ node bench.js 
lookup static route x 46,944,968 ops/sec ±0.50% (94 runs sampled)
lookup dynamic route x 3,446,964 ops/sec ±1.12% (93 runs sampled)
lookup dynamic multi-parametric route x 1,939,438 ops/sec ±1.35% (86 runs sampled)
lookup dynamic multi-parametric route with regex x 1,492,556 ops/sec ±0.73% (93 runs sampled)
lookup long static route x 3,396,039 ops/sec ±0.70% (92 runs sampled)
lookup long dynamic route x 2,359,074 ops/sec ±1.04% (95 runs sampled)
lookup static versioned route x 9,666,929 ops/sec ±0.51% (94 runs sampled)
find static route x 40,421,353 ops/sec ±0.30% (93 runs sampled)
find dynamic route x 4,331,766 ops/sec ±0.89% (92 runs sampled)
find dynamic multi-parametric route x 2,343,994 ops/sec ±1.23% (89 runs sampled)
find dynamic multi-parametric route with regex x 1,663,532 ops/sec ±0.28% (93 runs sampled)
find long static route x 5,093,791 ops/sec ±0.56% (89 runs sampled)
find long dynamic route x 3,361,727 ops/sec ±0.55% (92 runs sampled)
find static versioned route x 12,483,715 ops/sec ±0.73% (94 runs sampled)
$ 
```

It _seems_ like this is a better change performance-wise, and like it could potentially avoids some code duplication, so I thought I'd submit this in case that looks like an acceptable change.

### DIFF
--- a/index.js
+++ b/index.js
@@ -351,7 +351,6 @@ Router.prototype.find = function find (method, path, version) {
     path = path.toLowerCase()
   }
 
-  var encodedPath
   var maxParamLength = this.maxParamLength
   var currentNode = this.tree
   var wildcardNode = null
@@ -448,8 +447,48 @@ Router.prototype.find = function find (method, path, version) {
       pathLenWildcard = pathLen
     }
 
-    currentNode = node
+    // parametric route
+    if (kind === NODE_TYPES.PARAM) {
+      currentNode = node
+      i = path.indexOf('/')
+      if (i === -1) i = pathLen
+      if (i > maxParamLength) return null
+      decoded = fastDecode(originalPath.slice(idxInOriginalPath, idxInOriginalPath + i))
+      if (decoded === null) return null
+      params[pindex++] = decoded
+      path = path.slice(i)
+      idxInOriginalPath += i
+      continue
+    }
+
+    // wildcard route
+    if (kind === NODE_TYPES.MATCH_ALL) {
+      decoded = fastDecode(originalPath.slice(idxInOriginalPath))
+      if (decoded === null) return null
+      params[pindex] = decoded
+      currentNode = node
+      path = ''
+      continue
+    }
+
+    // parametric(regex) route
+    if (kind === NODE_TYPES.REGEX) {
+      currentNode = node
+      i = path.indexOf('/')
+      if (i === -1) i = pathLen
+      if (i > maxParamLength) return null
+      decoded = fastDecode(originalPath.slice(idxInOriginalPath, idxInOriginalPath + i))
+      if (decoded === null) return null
+      if (!node.regex.test(decoded)) return null
+      params[pindex++] = decoded
+      path = path.slice(i)
+      idxInOriginalPath += i
+      continue
+    }
+
+    // multiparametric route
     if (kind === NODE_TYPES.MULTI_PARAM) {
+      currentNode = node
       i = 0
       if (node.regex !== null) {
         var matchedParameter = path.match(node.regex)
@@ -459,21 +498,15 @@ Router.prototype.find = function find (method, path, version) {
         while (i < pathLen && path.charCodeAt(i) !== 47 && path.charCodeAt(i) !== 45) i++
         if (i > maxParamLength) return null
       }
-    } else if (kind === NODE_TYPES.MATCH_ALL) {
-      i = pathLen
-    } else {
-      i = path.indexOf('/')
+      decoded = fastDecode(originalPath.slice(idxInOriginalPath, idxInOriginalPath + i))
+      if (decoded === null) return null
+      params[pindex++] = decoded
+      path = path.slice(i)
+      idxInOriginalPath += i
+      continue
     }
 
-    if (i === -1) i = pathLen
-    if (i > maxParamLength) return null
-    encodedPath = originalPath.slice(idxInOriginalPath, idxInOriginalPath + i)
-    decoded = fastDecode(encodedPath)
-    if (decoded === null) return null
-    if (kind === NODE_TYPES.REGEX && !node.regex.test(decoded)) return null
-    params[pindex++] = decoded
-    path = path.slice(i)
-    idxInOriginalPath += i
+    wildcardNode = null
   }
 }
 
--- a/index.js
+++ b/index.js
@@ -340,24 +340,27 @@ Router.prototype.lookup = function lookup (req, res, ctx) {
 }
 
 Router.prototype.find = function find (method, path, version) {
-  if (this.caseSensitive === false) {
-    path = path.toLowerCase()
-  }
-
   if (path.charCodeAt(0) !== 47) { // 47 is '/'
     path = path.replace(FULL_PATH_REGEXP, '/')
   }
 
+  var originalPath = path
+  var originalPathLength = path.length
+
+  if (this.caseSensitive === false) {
+    path = path.toLowerCase()
+  }
+
+  var encodedPath
   var maxParamLength = this.maxParamLength
   var currentNode = this.tree
   var wildcardNode = null
   var pathLenWildcard = 0
-  var originalPath = path
-  var originalPathLength = path.length
   var decoded = null
   var pindex = 0
   var params = []
   var i = 0
+  var idxInOriginalPath = 0
 
   while (true) {
     var pathLen = path.length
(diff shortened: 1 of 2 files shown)


mcollina:
Thanks for keeping up with the work, this is perfect for me.

misterdjules:
New benchmark results with a minimum of 500 samples for the benchmark:

Running benchmark from master with node v10.14.1:

```
$ node bench.js
lookup static route x 48,625,148 ops/sec ±0.33% (590 runs sampled)
lookup dynamic route x 3,667,148 ops/sec ±0.40% (585 runs sampled)
lookup dynamic multi-parametric route x 2,025,568 ops/sec ±0.36% (589 runs sampled)
lookup dynamic multi-parametric route with regex x 1,482,397 ops/sec ±0.35% (591 runs sampled)
lookup long static route x 3,401,748 ops/sec ±0.29% (586 runs sampled)
lookup long dynamic route x 2,310,850 ops/sec ±0.34% (589 runs sampled)
lookup static versioned route x 8,931,153 ops/sec ±0.31% (587 runs sampled)
find static route x 35,781,234 ops/sec ±0.36% (583 runs sampled)
find dynamic route x 4,393,120 ops/sec ±0.41% (587 runs sampled)
find dynamic multi-parametric route x 2,296,578 ops/sec ±0.34% (589 runs sampled)
find dynamic multi-parametric route with regex x 1,667,617 ops/sec ±0.38% (591 runs sampled)
find long static route x 4,795,110 ops/sec ±0.35% (587 runs sampled)
find long dynamic route x 3,276,850 ops/sec ±0.37% (589 runs sampled)
find static versioned route x 10,636,877 ops/sec ±0.34% (589 runs sampled)
$ 
```

With the changes in this PR and same node version:

```
$ node bench.js 
lookup static route x 47,486,906 ops/sec ±0.38% (589 runs sampled)
lookup dynamic route x 3,503,609 ops/sec ±0.39% (586 runs sampled)
lookup dynamic multi-parametric route x 1,971,270 ops/sec ±0.33% (590 runs sampled)
lookup dynamic multi-parametric route with regex x 1,454,733 ops/sec ±0.42% (589 runs sampled)
lookup long static route x 3,396,126 ops/sec ±0.30% (588 runs sampled)
lookup long dynamic route x 2,387,300 ops/sec ±0.34% (590 runs sampled)
lookup static versioned route x 8,897,948 ops/sec ±0.26% (590 runs sampled)
find static route x 34,013,146 ops/sec ±0.57% (580 runs sampled)
find dynamic route x 4,284,053 ops/sec ±0.49% (589 runs sampled)
find dynamic multi-parametric route x 2,320,655 ops/sec ±0.35% (587 runs sampled)
find dynamic multi-parametric route with regex x 1,643,952 ops/sec ±0.42% (591 runs sampled)
find long static route x 4,829,306 ops/sec ±0.30% (588 runs sampled)
find long dynamic route x 3,383,730 ops/sec ±0.35% (590 runs sampled)
find static versioned route x 10,743,365 ops/sec ±0.43% (582 runs sampled)
$ 
```

So the performance impact is that performance seems to be worse by ~4.0% in the worst case (lookup of dynamic route), but is sometimes better by > 3% (find long dynamic route).

misterdjules:
I'll try another approach leveraging polymorphism/different prototypes to see if that yields better results in terms of performance.

misterdjules:
After trying different approaches to implement this change, I've seen differences between benchmark runs that were ran one after the other of ~5% for the code pulled from the tip of the master branch.

I'm not familiar with Benchmark.js and micro/synthetic benchmarks in general. I assume this could be due to a variety of reasons, including other processes on my development setup competing for resources.

So I'm actually not sure I can trust the benchmark results for looking into differences that are < 5% even when using 500 or 1000 samples for benchmark runs. It makes determining whether this PR introduces a performance regression quite challenging. 

Looking at profiling data from V8's profiler, I was not able to identify any outstanding difference in terms of performance between current master and the changes in this PR.

I wanted to take a look at the code generated by V8, but failed to do that. This could be due to the fact that I'm not familiar with how to do that with recent versions of V8. For instance I was not able to look at the generated code for the `Router.prototype.find` function.

Let me know if you have suggestions on how to get more reliable results from benchmarks or how to look at the generated code, but otherwise I wouldn't think that the performance impact of these changes are a blocker.

mcollina:
I’ll try running them on a dedicated server to see if I can get stable numbers, but this code is ok for me and we can land.

mcollina:
I can confirm there is a slight drop. Using a dedicated servers shows a consistent drop of around 2% in the "lookup static route" benchmark. The rest are more or less in line (or harder to validate).

IMHO we should land this.

misterdjules:
@mcollina 

> Using a dedicated servers shows a consistent drop of around 2% in the "lookup static route" benchmark.

Do you generally use a specific setup to run those micro benchmarks? Using a laptop seems like it would be challenging to get consistent performance.

mcollina:
I have a dedicated server for 35€/month that I use for the most CPU intensive development. It’s cheaper than buying a workstation or a top of the line macbook pro.

delvedor:
Not sure if this should be landed as patch, minor or major.
All of them are both wrong and correct at the same time!

misterdjules:
@delvedor I agree one could advocate for any of them 😂 One could argue that it's better to be conservative when it comes to potentially breaking customers. The downside is that if you maintain only one release line you're basically forcing customers to adopt next major to grab future minor/patch changes.

But at least customers opt into that potentially breaking change deliberately and explicitly, as opposed to without knowing it.

So I would advocate for a major bump, but I don't have strong feelings either way :)

delvedor:
In the end, I think this is a breaking change. It is better to be safe and communicate clearly with users.

But before land this, I found a way to avoid a breaking change, and land this as a feature release.
If we add a new option, `keepCase` with a default value of `false`, we can maintain the current behavior and enhance it if you need to keep the letter casing.
I already have a working implementation, if you want I can push in your branch so you can see how it works.

misterdjules:
@delvedor 

> If we add a new option, `keepCase` with a default value of `false`, we can maintain the current behavior and enhance it if you need to keep the letter casing.

That sounds good to me! 

> I already have a working implementation, if you want I can push in your branch so you can see how it works.

That would be great, thank you!

delvedor:
Hello! I thought about this a lot, and I've finally taken a decision.
The current behavior of `caseSensistive` is wrong because we are changing the value that we are passing to the user. This pr should be a bug fix, but there may be someone that relies on this behavior, so we should cut a major.
Adding the `keepCase` would fix the problem, but it's just a monkey patch, at this point it makes more sense fix it once for all.

I'll update the README after merging this.
Thank you a lot for all the good work! :)

PR #116, opened 2019-02-13
increase min samples for benchmark suite

misterdjules:
See the added comment for context. This change results in the benchmark suite taking a lot longer to run, but I find the tradeoff acceptable in my recent use cases.

I don't feel super strongly about this as I can always override this locally, but maybe that's a better default value for everyone?

### DIFF
--- a/bench.js
+++ b/bench.js
@@ -1,6 +1,14 @@
 'use strict'
 
 const Benchmark = require('benchmark')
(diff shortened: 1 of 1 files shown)


mcollina:
Does this impact the other PR?

misterdjules:
It does, let me share the benchmark results there.

PR #117, opened 2019-02-14
fix error under --disallow-code-generation-from-strings

ethanresnick:
Hi, 

I'm trying to use find-my-way in a security-critical environment where I have to run node with the --disallow-code-generation-from-strings flag. Under this flag, calls to `new Function` error, preventing me from using find-my-way.

This PR adds a fallback approach in the `buildHandlers` function for when `new Function` errors. Because `buildHandlers` is only called once, having this fallback has essentially zero performance impact on people running in contexts where `new Function` is allowed. (They get the exact same handlers functions they were getting before.)

For the few people running under the --disallow-code-generation-from-strings flag, the fallback approach is certainly slower than the eval approach (it brings down the library's overall performance by ~8% on the router-benchmarks suite), but having code that works a bit slower seems far better than blowing up entirely.

### DIFF
--- a/node.js
+++ b/node.js
@@ -234,7 +234,18 @@ function buildHandlers (handlers) {
     code += `this['${m}'] = handlers['${m}'] || null
     `
   }
-  return new Function('handlers', code) // eslint-disable-line
+
+  try {
+    return new Function('handlers', code) // eslint-disable-line
+  } catch(e) {
+    return function(_handlers) {
+      _handlers = _handlers || {};
+      for (var i = 0; i < http.METHODS.length; i++) {
+        var m = http.METHODS[i];
+        this[m] = _handlers[m] || null;
+      }
+    };
+  }
 }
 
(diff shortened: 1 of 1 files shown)


ethanresnick:
@mcollina Is that comment directed at me? If so, I think it'd be a great idea to run the test suite with that flag enabled too, to makes sure the behavior matches and that there aren't changes to the library in the future that cause it to stop working under that option. I'm not very familiar with Travis, though, so I'm probably the wrong person to set that up.

mcollina:
Closing for no activity, I might have lost this. In case this is still relevant, feel free to open a new PR.

ethanresnick:
@mcollina I’m happy to open a new PR and rebase the changes as necessary if someone else can tackle the Travis side of things, or if this can be merged without that. I don’t really have the bandwidth to dig into the CI setup. I do think this is still relevant though

PR #118, opened 2019-02-18
Use safe-regex2

mcollina:
https://snyk.io/test/github/fastify/fastify


delvedor:
For posterity: The new version of `safe-regex` has a security vulnerability, we have forked it, and we will maintain it. We couldn't use the previous version of the library since it was no longer maintaned.

See also https://github.com/delvedor/find-my-way/pull/100 and https://github.com/fastify/fastify/pull/1453 for context.

PR #130, opened 2019-08-19
fix prettyPrint indent

### LINKED ISSUE #129: prettyPrint outputs wrong indent
I found a glitch with `prettyPrint`.
Take the following code as an example:
```js
fastify.get('/pet/:id', async function (req, rep) {})
fastify.post('/pet/:id', async function (req, rep) {})
fastify.get('/preview', async function (req, rep) {})
```

It will finally output like this, where the indent of the route `post('/pet/:id')` is not correct:
S9U~X@(K2%{EX AYE}IG9L

I know that `prettyPrint` is depreciated, but it seems just a small bug that can be easily fixed.

fralonra:
fix #129

It outputs the following:
5P10B82%$6S$32QNT @61SY

### DIFF
--- a/node.js
+++ b/node.js
@@ -203,7 +203,7 @@ Node.prototype.prettyPrint = function (prefix, tail) {
           paramName += param + ` (${method})\n`
           return
         }
-        paramName += '    ' + prefix + ':' + param + ` (${method})`
+        paramName += prefix + '    :' + param + ` (${method})`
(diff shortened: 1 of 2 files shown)


mcollina:
Would you mind adding a unit test? https://github.com/delvedor/find-my-way/blob/master/test/pretty-print.test.js

fralonra:
A unit test has been added!

PR #136, opened 2019-10-08
Fix uppercase static path

### LINKED ISSUE #135: Multiple route with multiple path param and case insensitive
Hi,
this snippet:

```js
const http = require('http')
const router = require('find-my-way')({
  caseSensitive: false
})

router.on('POST', '/:param/User/:userId/Save', (req, res, params) => { res.end('{"message":"hello world"}') })
router.on('POST', '/:param/User/:userId/Get', (req, res, params) => { res.end('{"message":"hello world"}') })

  router.lookup(req, res)
})

server.listen(3000, err => {
  if (err) throw err
  console.log('Server listening on: http://localhost:3000')
})
```

will throw:

```
assert.js:373
    throw err;
    ^

AssertionError [ERR_ASSERTION]: There is already a child with label ':'
    at Node.addChild (\node_modules\find-my-way\node.js:58:3)
```

If I set only one route, all works as expected.

Do you have any suggestion to solve?
If this is a bug could you suggest me where to start looking to help?

Eomm:
Fixes #135

### DIFF
--- a/test/case-insensitive.test.js
+++ b/test/case-insensitive.test.js
@@ -201,3 +201,26 @@ test('case insensitive with wildcard', t => {
 
   findMyWay.lookup({ method: 'GET', url: '/FOO/baR', headers: {} }, null)
 })
+
+test('parametric case insensitive with multiple routes', t => {
+  t.plan(4)
+
+  const findMyWay = FindMyWay({
+    caseSensitive: false,
+    defaultRoute: (req, res) => {
+      t.fail('Should not be defaultRoute')
+    }
+  })
+
+  findMyWay.on('POST', '/foo/:param/Static/:userId/Save', (req, res, params) => {
+    t.equal(params.param, 'bAR')
+    t.equal(params.userId, 'one')
+  })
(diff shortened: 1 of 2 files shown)


test/case-insensitive.test.js
```diff
@@ -201,3 +201,26 @@ test('case insensitive with wildcard', t => {
...
+    t.equal(params.param, 'bAR')
+    t.equal(params.userId, 'one')
+  })
+  findMyWay.on('POST', '/foo/:param/Static/:userId/Update', (req, res, params) => {
+    t.equal(params.param, 'Bar')
+    t.equal(params.userId, 'two')
+  })
+
+  findMyWay.lookup({ method: 'POST', url: '/foo/bAR/static/one/SAVE', headers: {} }, null)
+  findMyWay.lookup({ method: 'POST', url: '/foo/Bar/static/two/update', headers: {} }, null)
```
mcollina: Would you mind adding one more lookup with some case changes in both `foo` and `static`?

Eomm: Done

mcollina:
Good work!

PR #146, opened 2020-02-21
No match path param with ignoreTrailingSlash

### LINKED ISSUE #145: No match when ignoreTrailingSlash is set to true
## Bug Report
This happens when `ignoreTrailingSlash` option is set to `true`

```js
const router = require("find-my-way")({ ignoreTrailingSlash: true });

router.get('/a/c', () => {});
router.get('/a/:whatever/b', () => {});

router.find('GET', '/a/c/b');
// null
router.find('GET', '/a/AnythingExcept"c"/b');
// {handler: Function, params: {whatever: "AnythingExcept\"c\""}, store: null}
```

## Expected behavior

Path that must match:
`/a/AnythingHere/b`

Path that does not get a match:
`/a/c/b`

## Environment

- *node version*: 12.14.1
- *find-my-way version*: 2.2.1
- *os*: Ubuntu 18.04

Eomm:
Fixes #145 

Still working on it

### DIFF
--- a/test/issue-145.test.js
+++ b/test/issue-145.test.js
@@ -4,7 +4,7 @@ const t = require('tap')
 const FindMyWay = require('../')
 
 t.test('issue-145', (t) => {
-  t.plan(4)
+  t.plan(8)
 
   const findMyWay = FindMyWay({ ignoreTrailingSlash: true })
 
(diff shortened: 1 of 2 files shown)


test/issue-145.test.js
```diff
@@ -0,0 +1,22 @@
...
+
+const noop = function () {}
+
+t.test('issue-145', (t) => {
+  t.plan(4)
+
+  const findMyWay = FindMyWay({ ignoreTrailingSlash: true })
+
+  findMyWay.on('GET', '/a/b', noop)
+  findMyWay.on('GET', '/a/:pam/c', noop)
```
mcollina: Instead of using noop, can you please use named functions so that we can assert the right one matched?

PR #152, opened 2020-05-17
fix #151 Wildcard route should not be blocked by Parametric with diff…

### LINKED ISSUE #151: Wildcard bug
```js
const http = require('http');
const router = require('find-my-way')();

router.on('OPTIONS', '/*', (req, res, params) => {
  res.end('{"hello": "world"}');
});
router.on('OPTIONS', '/obj/*', (req, res, params) => {
  res.end('{"hello": "wildcard"}');
});
router.on('GET', '/obj/:id', (req, res, params) => {
  res.end('{"hello": "paramtric"}');
});

  router.lookup(req, res);
});

server.listen(3000, err => {
  if (err) throw err;
  console.log('Server listening on: http://localhost:3000');
  // test
  const req = http.request({
    host: 'localhost',
    port: 3000,
    method: 'OPTIONS',
    path: '/obj/params'
  }, res => {
    const chunks = [];
    res.on('data', chunk => chunks.push(chunk));
    res.on('end', () => {
      const data = Buffer.concat(chunks).toString();
      console.log(data); // result string 
    });
  });
  req.end();
});
```

As I expect，the result `data` should be `{"hello": "wildcard"}`.

Actually, it is `{"hello": "world"}`.

muhonglong:
demo at #151

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,7 +18,7 @@ const isRegexSafe = require('safe-regex2')
 const Node = require('./node')
 const NODE_TYPES = Node.prototype.types
 const httpMethods = http.METHODS
-const FULL_PATH_REGEXP = /^https?:\/\/.*\//
+const FULL_PATH_REGEXP = /^https?:\/\/.*?\//
 
 if (!isRegexSafe(FULL_PATH_REGEXP)) {
   throw new Error('the FULL_PATH_REGEXP is not safe, update this module')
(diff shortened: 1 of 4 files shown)


muhonglong:
Another problem found:
```js
router.find('GET', 'https://localhost/a/b/c')
// eql
router.find('GET', '/c')
// expect
router.find('GET', '/a/b/c')
```

mcollina:
find-my-way does not support full urls.

delvedor:
Landed in `v3.0.1` and `v2.2.3` 🎉

Issue #162, opened 2020-09-08
sanitize on #find


spasquali-ww:
Hi, 

Version 3.0.1

When using [lookup()](https://github.com/delvedor/find-my-way/blob/ce6d38b8a9430048934ddadb74a33950e7344101/index.js#L352) I see that the `req.url` is passed through [sanitizeUrl()](https://github.com/delvedor/find-my-way/blob/ce6d38b8a9430048934ddadb74a33950e7344101/index.js#L353) prior to forwarding the request to [find()](https://github.com/delvedor/find-my-way/blob/ce6d38b8a9430048934ddadb74a33950e7344101/index.js#L360)

We're calling `find()` directly. `find()` does not run [sanitizeUrl()](https://github.com/delvedor/find-my-way/blob/ce6d38b8a9430048934ddadb74a33950e7344101/index.js#L610) on the `path`(`req.url`) it receives, which had caused some issues on our end.

Now we are simply running our own copy of `sanitizeUrl` prior to calling `find()` but are wondering if `find-my-way` should do this itself. Since `lookup()` always forwards to `find()`, and `sanitizeUrl()` is not used/referenced anywhere else in `lookup()`, it seems that removing the `sanitizeUrl(req.url)` from `lookup()` and transplanting it to `find()` would be simplest.

Thanks for the great router!

PR #163, opened 2020-09-08
refactor: move sanitizeUrl() to find(); remove from lookup())

spasquali-ww:
Referencing: https://github.com/delvedor/find-my-way/issues/162#issue-695999569

### DIFF
--- a/index.js
+++ b/index.js
@@ -358,8 +358,7 @@ Router.prototype.lookup = function lookup (req, res, ctx) {
 }
 
 Router.prototype.find = function find (method, path, version) {
-
-  path = sanitizeUrl(path);
+  path = sanitizeUrl(path)
 
   if (path.charCodeAt(0) !== 47) { // 47 is '/'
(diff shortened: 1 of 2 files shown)


Eomm:
Could you add some test to cover your issue and avoid regressions in future?

delvedor (changes requested):
Hello! This behavior is intended.
`find` should only look if something has been stored under that path.
The reason behind it is because you can also use find my way as a store for something else than routes, and sanitizing might be a problem.
`lookup` is the api for working with http.

spasquali-ww:
Aha. Please close this then. Sorry for the confusion!

Intrigued about using `find()` for storing other 'things'...

delvedor:
Thank you for contributing!

Yeah, the fourth parameter in a route declaration is a store object, where you can put anything you like and you can retrieve it with `find`.
Ma his is also how we are handling the context of each route in Fastify :)

PR #167, opened 2020-10-14
Support optional parameters at the end of the path

JulianBerger:
Originally wanted to add this feature to fastify: https://github.com/fastify/fastify/pull/2607 but we came to the conclusion that it fits better into here.

Support optional parameters (with a '?' suffix) if they are at the end of the path.

Example: 
**Input Route:** `/v1/user/:userId/admin/mail/:id?`
Can be accessed by:

1. `/v1/user/:userId/admin/mail/:id`
2. `/v1/user/:userId/admin/mail`

More information about the regex:
Screen Shot 2020-10-14 at 16 17 07

Screen Shot 2020-10-14 at 16 17 21
So the trailing slash is still preserved with the second capture group.

### DIFF
--- a/index.js
+++ b/index.js
@@ -70,6 +70,17 @@ Router.prototype.on = function on (method, path, opts, handler, store) {
   // handler validation
   assert(typeof handler === 'function', 'Handler should be a function')
 
+  // path ends with optional parameter
+  if (path.endsWith('?') || path.endsWith('?/')) {
+    const optionalParamRegex = /(\/:[^/]*?)\?(\/?)$/
+    const pathFull = path.replace(optionalParamRegex, '$1')
+    const pathOptional = path.replace(optionalParamRegex, '$2')
+
+    this.on(method, pathFull, opts, handler, store)
+    this.on(method, pathOptional, opts, handler, store)
+    return
+  }
+
   this._on(method, path, opts, handler, store)
 
   if (this.ignoreTrailingSlash && path !== '/' && !path.endsWith('*')) {
--- a/index.js
+++ b/index.js
@@ -331,6 +331,17 @@ Router.prototype.off = function off (method, path) {
   assert(path.length > 0, 'The path could not be empty')
   assert(path[0] === '/' || path[0] === '*', 'The first character of a path should be `/` or `*`')
 
+  // path ends with optional parameter
+  const optionalParamRegex = /(\/:[^/]*?)\?(\/?)$/
+  if (path.match(optionalParamRegex)) {
+    const pathFull = path.replace(optionalParamRegex, '$1$2')
+    const pathOptional = path.replace(optionalParamRegex, '$2')
+
+    this.off(method, pathFull)
+    this.off(method, pathOptional)
+    return
+  }
+
   // Rebuild tree without the specific route
   const ignoreTrailingSlash = this.ignoreTrailingSlash
   var newRoutes = self.routes.filter(function (route) {
--- a/index.js
+++ b/index.js
@@ -71,8 +71,8 @@ Router.prototype.on = function on (method, path, opts, handler, store) {
   assert(typeof handler === 'function', 'Handler should be a function')
 
   // path ends with optional parameter
-  if (path.endsWith('?') || path.endsWith('?/')) {
-    const optionalParamRegex = /(\/:[^/]*?)\?(\/?)$/
+  const optionalParamRegex = /(\/:[^/]*?)\?(\/?)$/
+  if (path.match(optionalParamRegex)) {
     const pathFull = path.replace(optionalParamRegex, '$1$2')
     const pathOptional = path.replace(optionalParamRegex, '$2')
 
--- a/index.js
+++ b/index.js
@@ -19,9 +19,14 @@ const Node = require('./node')
 const NODE_TYPES = Node.prototype.types
 const httpMethods = http.METHODS
 const FULL_PATH_REGEXP = /^https?:\/\/.*?\//
+const OPTIONAL_PARAM_REGEXP = /(\/:[^/()]*?)\?(\/?)/
 
 if (!isRegexSafe(FULL_PATH_REGEXP)) {
   throw new Error('the FULL_PATH_REGEXP is not safe, update this module')
+}
+
+if (!isRegexSafe(OPTIONAL_PARAM_REGEXP)) {
+  throw new Error('the OPTIONAL_PARAM_REGEXP is not safe, update this module')
 }
 
 const acceptVersionStrategy = require('./lib/accept-version')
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -70,6 +70,17 @@ Router.prototype.on = function on (method, path, opts, handler, store) {
   // handler validation
   assert(typeof handler === 'function', 'Handler should be a function')
 
+  // path ends with optional parameter
+  const optionalParamRegex = /(\/:[^/]*?)\?(\/?)$/
```
delvedor, line 74: Out of curiosity, why the need for a regex instead of `path.endsWith('?')`

JulianBerger: I had `path.endsWith('?') || path.endsWith('?/')` before, but then it's not ensured to be a parameter.
That would also match for routes like `/foo?`, don't know if that is an issue..

delvedor: Gotcha! 👍

delvedor (changes requested):
Can you add some additional test cases?
Eg:
- what happens if someone declares two following routes?
  - `/foo/:bar?`
  - `/foo`
- what happens if someone puts the question mark inside a route declaration?
  - `/foo/:bar?/baz`

It's useful to test also "bad behaviors", so we can be sure not to break anything in the future.

mcollina (approved):
lgtm

test/optional-params.test.js
```diff
@@ -0,0 +1,98 @@
...
+
+test('Test for param with ? not at the end', (t) => {
+  t.plan(1)
+  const findMyWay = FindMyWay({
+    defaultRoute: (req, res) => {
+      t.fail('Should not be defaultRoute')
+    }
+  })
+
+  findMyWay.on('GET', '/foo/:bar?/baz', (req, res, params) => {
```
delvedor, line 53: Given that this case is not supported, I think we should throw an error here, what do you think?

index.js
Eomm, line 74: moving this regexp near to `FULL_PATH_REGEXP` would avoid to recreate it for every route/`on` call. It should optimize a little bit the GC execution + remove the duplicated declaration*

index.js
```diff
@@ -320,6 +331,17 @@ Router.prototype.off = function off (method, path) {
   assert(path.length > 0, 'The path could not be empty')
   assert(path[0] === '/' || path[0] === '*', 'The first character of a path should be `/` or `*`')
 
+  // path ends with optional parameter
+  const optionalParamRegex = /(\/:[^/]*?)\?(\/?)$/
```
Eomm, line 335: the duplicated declaration*

index.js
mcollina: It would also avoid future bugs.

test/optional-params.test.js
JulianBerger: Yes, it's throwing an Error now.

index.js
JulianBerger: Good idea, declaring it globally now.

JulianBerger (commented):
All resolved

mcollina (approved):
lgtm

JulianBerger:
@delvedor can you merge it please? :)

delvedor (changes requested):
Hello! Sorry for the delay :S
Everything looks good to me, can you add one last test to verify the behavior with `ignoreTrailingSlash: true`?

JulianBerger:
Added tests for ignoreTrailingSlash.

test/optional-params.test.js
```diff
@@ -0,0 +1,144 @@
...
+
+  findMyWay.on('GET', '/a/:p1-:p2?', (req, res, params) => {
+    if (params.p1 && params.p2) {
+      t.equal(params.p1, 'foo')
+      t.equal(params.p2, 'bar-baz')
+    }
+  })
+
+  findMyWay.lookup({ method: 'GET', url: '/a/foo-bar-baz', headers: {} }, null)
+  // findMyWay.lookup({ method: 'GET', url: '/a/foo', headers: {} }, null)
```
Eomm, line 77:
```suggestion
```

test/optional-params.test.js
```diff
@@ -0,0 +1,144 @@
...
+  findMyWay.lookup({ method: 'GET', url: '/test/hello/foo', headers: {} }, null)
+  findMyWay.lookup({ method: 'GET', url: '/test/hello/foo/', headers: {} }, null)
+})
+
+test('Optional Parameter with ignoreTrailingSlash = false', (t) => {
+  t.plan(4)
+  const findMyWay = FindMyWay({
+    ignoreTrailingSlash: false,
+    defaultRoute: (req, res) => {
+      t.notMatch(req.url, '.*/$') // urls without trailing slash
```
Eomm, line 109:
```suggestion
      t.fail('Should not be defaultRoute')
```

test/optional-params.test.js
```diff
@@ -0,0 +1,144 @@
...
+test('Optional Parameter with ignoreTrailingSlash = true', (t) => {
+  t.plan(4)
+  const findMyWay = FindMyWay({
+    ignoreTrailingSlash: true,
+    defaultRoute: (req, res) => {
+      t.fail('Should not be defaultRoute')
+    }
+  })
+
+  findMyWay.on('GET', '/test/hello/:optional?/', (req, res, params) => {
```
delvedor, line 90: There is no need to add the trailing slash in the route definition, it might be passed by the client, which you are testing correctly.

```suggestion
  findMyWay.on('GET', '/test/hello/:optional?', (req, res, params) => {
```

test/optional-params.test.js
```diff
@@ -0,0 +1,144 @@
...
+test('Optional Parameter with ignoreTrailingSlash = false', (t) => {
+  t.plan(4)
+  const findMyWay = FindMyWay({
+    ignoreTrailingSlash: false,
+    defaultRoute: (req, res) => {
+      t.notMatch(req.url, '.*/$') // urls without trailing slash
+    }
+  })
+
+  findMyWay.on('GET', '/test/hello/:optional?/', (req, res, params) => {
```
delvedor, line 113:
```suggestion
  findMyWay.on('GET', '/test/hello/:optional?', (req, res, params) => {
```

delvedor (changes requested):
Great work, one little nit, you should not register the route with the trailing slash, it should be used only in the "request".

JulianBerger:
> Great work, one little nit, you should not register the route with the trailing slash, it should be used only in the "request".

Ahh sure, fixed that! Also changed the ignoreTrailingSlash = false test behavior: `/test/hello/` will give us an empty string as the `:optional` parameter...

test/optional-params.test.js
```diff
@@ -0,0 +1,144 @@
...
+  t.plan(4)
+  const findMyWay = FindMyWay({
+    ignoreTrailingSlash: false,
+    defaultRoute: (req, res) => {
+      t.match(req.url, '/test/hello/foo/')
+    }
+  })
+
+  findMyWay.on('GET', '/test/hello/:optional?', (req, res, params) => {
+    console.log('req', req, 'opt', params.optional)
```
delvedor, line 113:
```suggestion
```

test/optional-params.test.js
```diff
@@ -0,0 +1,144 @@
...
+    if (params.optional) {
```
delvedor, line 114: This test looks brittle, can you write a more strict if check?

delvedor:
@JulianBerger Hello! Do you have time to finish this pr? :)

mcollina:
Closing for no activity

jcward:
FYI for anyone who lands here looking for optional trailing parameters in find-my-way -- it suited my needs to simply register the same handler twice:

```
    findMyWay.on('GET', '/list/:id', handle_list_items);
    findMyWay.on('GET', '/list', handle_list_items);
```

PR #170, opened 2020-11-02
Constrained routes take 2

### LINKED ISSUE #165: Feature discussion: Support for request-based constraints
## 1. Overview:

Find-my-way currently supports constraining routes only by versions. 

The following is a proposal, based on the discussion in Fastify's issue [#2498](https://github.com/fastify/fastify/issues/2498), to add support for more generic request-based constraints.

@airhorns already made the case for both the motivation and utility of this feature, so I'll be going over the technical implementation details directly.

### A) Considerations:

To start off, these are some important things that need to be taken into account, as outlined in the issue referenced above:

1. The route should be fully decided by the router
2. The suggested solution shouldn't negatively affect the performance of:
    - Non-constrained routes
    - Version-only constrained route
    - Other constrained routes (to some extent, will detail below what I mean exactly)
3. Eventhough the focus of that issue was host-based constraining, it would be great if there is a generic way for request-based constraining that doesn't affect performance

### B) Analysis of relevant find-my-way concepts:

Currently, find-my-way uses internally a highly performant Radix Tree in order to store and route requests. 

Each node of the tree (that represents a supported path) stores a `handlers` object that maps each supported http method for that path to its handler.

Additionally, versioning routes is allowed by using a default accept-version strategy, *that can be overriden by the user at their own risk*.

This strategy contains two properties:

- `storage` - the factory function for the Storage of the handlers based on their version, it returns an instance of the `semver-store`.
- `deriveVersion` - the function to determine the version based on the request

For versioned routes, a `storage` instance is created for each node and used to map each supported http method for that path to its handler, instead of the `handlers` object used for non-versioned routes.

### C) Solution:

In RoR it is possible to constrain a route based on any method on the request object that returns a string, as detailed [here](https://guides.rubyonrails.org/routing.html#request-based-constraints):

> Request constraints work by calling a method on the [Request object](https://guides.rubyonrails.org/action_controller_overview.html#the-request-object) with the same name as the hash key and then compare the return value with the hash value. Therefore, constraint values should match the corresponding Request object method return type. For example: constraints: { subdomain: 'api' } will match an api subdomain as expected, however using a symbol constraints: { subdomain: :api } will not, because request.subdomain returns 'api' as a String.

This means that the user has the ability to use any combination of constraints, where each one can be:

1. Equal to the corresponding request value
2. A regex that matches to the corresponding request value
3. An [advanced constraint](https://guides.rubyonrails.org/routing.html#advanced-constraints) where the user provides an object that responds to **matches? method** that Rails should use.
    - @mcollina mentioned [here](https://github.com/fastify/fastify/issues/2498#issuecomment-677703974) that this option isn't interesting as a feature, because the route needs to be fully decided by the router itself.

With this as a starting point, my suggested solution to allow support for additional constraints in a generic way is to use at the router level a new `acceptConstraintsStrategy` object instead of the current `acceptVersionStrategy`. 

The object would maintain the same structure:

- `storage` - the factory function for the Storage of the handlers based on their constraints, it returns an instance of a new `ConstraintsStore`; this store can contain child strategies (version, host, etc...) that will each have the same structure (`storage` and `deriveConstraint`).
- `deriveConstraints` -  the function to determine the constraints values based on the request

The `ConstraintsStore` will expose methods `set`, `get`, and `empty`, and will match constraints following this logic:

As an example, let's take these constraints: 

```jsx
{ version: '1.0.x', host: 'auth.airhorns.dev' }
```

- Calling the `set` method with these constraints will result in a call to the `set` method of all the strategies used, in this case `version` and `host`
- Calling the `get` method will result in a call to the `get` method of all the strategies used, in this case `version` and `host`. Since the strategies store handlers independently, if one of them returns `null` then we can know that no match that fulfills all constraints is found and we return `null`, otherwise we return the stored handlers.

Addtionally, given the analysis in the previous section, it's important to note that the use of versioned routes storage is only done if `deriveVersion` returns a non empty value.

This works well for the version constraint, which has to be explicitly set in the `accept-version` header. For other constraints like the header `host`, it will always contain a value which would imply the constrained storage will always be used (Remember that each router tree node currently contains separate non-versioned and versioned storage handlers). 

This raises a few questions that I'd like to open a discussion about:

1. (Somewhat off-topic) In the current code, if the user provides an `accept-version` header with value `*` (which allows matching any version), but no version is specified when registering the route. This would return no handler since the versioned handler storage would be used. Is this by design?

2. Should we support constraints for all request object properties by default, similar to RoR? 
3. If default support is added, how do you propose matching be done? Given the importance of performance for fastify, I'd assume use of regex matching is out of question, that leaves two options as far as I can tell:
    - Exact match
    - Default strategies provided for all request properties
4. How should the router api be updated to use constraints? Have a `constraints` property inside the `opts` parameter when registering the route? Replace the `opts` param with `constraints`? Or do you have a better suggestion?
5. If you agree with my reasoning, what would be the best way to allow the user to provide custom strategies?

airhorns:
This is about https://github.com/fastify/fastify/issues/2498 (host based routing), closes #165, replaces and closes #166. This is a completed version of @AyoubElk's changes in #166 that passes the tests!

This: 
 - keeps support for semver based version constrained routes
 - keeps the old behaviour of not matching a versioned route if no `Accept-Version` route is passed, which requires lots of code
 - adds support for host constrained routes that only match if the `Host` header is equal to a string or passes a regexp test
 - adds support for arbitrary constraints passed as objects that know how to inspect a request and decide if a handler should match for that request
 - adds support for testing multiple constraints for a handler, and matching only handlers who pass all the constraints

Implementation details:
 - `find` now (must) take an object of constraint values derived from the request. Before, it took an optional version, but now there are many values, so it takes an object with a `version` key and or others. This is a weird API if you're calling `find` directly, but invisible if you're calling `lookup`, which is what Fastify itself does. This is done to avoid extra checks and allocations for every route match. It also generates a lot of noise in this PR, sorry.
 - Nodes now know how to split themselves
 - Nodes no longer have a `.handler` and a `.versionStore` that store handlers separately, they have an array of handlers, and sometimes many constraint stores that know how to report which handlers should match which constraint values. The old version store is one, the host store is another, and userland code can create more.
 - There's a `Constrainer` object that holds all the constraining strategies and that knows how to validate constraints
 - Constraint testing uses fast and memory efficient bitwise operations to figure out which handlers match all the constraints (when there are constraints for a route). This is tricky but necessary because we need to return handlers that match *all* the constraints, not just some of them, so we need to keep some intermediate state when checking about what handlers have passed all the constraints so far. 
 - An optimized function to get the `derivedConstraints`, which is the request host and the `accept-version` header most of the time is `new Function()`'d for speed since it is run for every request match
 - An optimized constraint matcher function using the bitmapping algorithm described in the comments is `new Function()`'d for each handler that has constraints, which creates a megamorphic callsite but only for constrainted handlers and in my testing is still faster than a monomorphic callsite with a loop

Notably, there's a huge performance regression right now, especially on static route matching. First make it work, then make it fast they say, which I have been trying but failing to do, so I'd really love any assistance or even just theories about what is causing a big slow down. Currently a Draft PR until performance is brought back in line. 

Would love your feedback on the code and specifically any pointers for where I screwed up the performance! On master I see roughly 45M static route matches a second, and I see 32M on this branch, so there's a lot of work yet to do. 

The remaining work: 
 - [x] Aforementioned performance improvement
 - [x] CHANGELOG entry and upgrade info. The only breaking change would be to the shape of custom version strategies, which I could potentially get rid of if strictly necessary at the expense of some nasty code. 

Also just FYI @AyoubElk and I have been coordinating on this off platform, I stepped in to complete the work so please direct your feedback on this feature to us here!


lib/strategies/accept-host.js
```diff
@@ -0,0 +1,40 @@
...
+}
+
+module.exports = {
+  name: 'host',
+  mustMatchWhenDerived: false,
+  storage: Hostvalue,
+  validate (value) {
+    assert(typeof value === 'string' || Object.prototype.toString.call(value) === '[object RegExp]', 'Host should be a string or a RegExp')
+  }
+}
```
airhorns, line 40: Someday, this could use a radix tree too, but for simplicity it's just a regex match for now. #patcheswelcome

mcollina:
good work!

index.js
```diff
@@ -397,9 +368,7 @@ Router.prototype.find = function find (method, path, version) {
     var previousPath = path
     // found the route
     if (pathLen === 0 || path === prefix) {
-      var handle = version === undefined
-        ? currentNode.handler
-        : currentNode.getVersionHandler(version)
+      var handle = currentNode.getMatchingHandler(derivedConstraints)
```
mcollina, line 371: you lose throughput here, this function call was not in the hot path before.

node.js
```diff
@@ -96,98 +99,211 @@ Node.prototype.addChild = function (node) {
...
+
 Node.prototype.findByLabel = function (path) {
   return this.children[path[0]]
 }
 
-Node.prototype.findChild = function (path) {
+Node.prototype.findMatchingChild = function (derivedConstraints, path) {
   var child = this.children[path[0]]
-  if (child !== undefined && (child.numberOfChildren > 0 || child.handler !== null)) {
+  if (child !== undefined && (child.numberOfChildren > 0 || child.getMatchingHandler(derivedConstraints) !== null)) {
```
mcollina, line 143: you are calling getMatchingHandler 3 times in this function, this is probably superflous.

package.json
```diff
@@ -44,6 +44,7 @@
   },
   "dependencies": {
     "fast-decode-uri-component": "^1.0.1",
+    "fast-deep-equal": "^3.1.3",
```
mcollina, line 47: where is this called?

node.js
airhorns: The `child` changes for each call, it's the piece that implements static > param > wildcard route checking

package.json
airhorns: Used for checking if duplicate constraints are passed for the same path. It used to be that you could just check `version == handler.version`, but now that there can be more than one constraint, it has to be some by-value comparison. Happy to vendor an implementation of object equality by value, but I couldn't think of a way to avoid the check so I figured i'd use someone else's maintained code.

index.js
airhorns: Indeed, but weirdly, on @AyoubElk 's branch he also switched this to making a function call every time and had the static routing performance inline with `master`. It also seems like one more monomorphic function call shouldn't cause a 30% regression? I also tried commenting out the function call and returning just `.handlers[0]` to mimic the old behaviour just to test performance, and performance didn't really recover much. So, I feel like I'm missing something :/

airhorns:
Ok, after some discussion with @AyoubElk , I think we ended up in an ok place. After the most recent updates:

 - the constraint system isn't used until the first time a route is constrained, which preserves the new behaviour in #172 
 - static route matching is slightly faster than the master branch for routers which don't have constraints
 - static route matching is slower than the master branch for routers which do have constraints

The constraining code pretty much has to do some object allocation that wasn't present before in order to support multiple constraints. Before this branch, deriving the version for a request was returning a reference to a string already in memory, but now with multiple constraints, the most straightforward way to represent all the derived values is to use an object. Allocating that object for every route match was the main thing throttling the benchmarks above. So, now we do that only if some routes need to actually look at the value, which makes route matching *faster* if you aren't using versioned routes anywhere, because we skip deriving even the version. So that's good! 

The problem with this approach is that as soon as you add one version'd or host constrain'd route, that object starts getting allocated once per route match. Peak performance is maintained but for complicated routing setups we have one (1) more object allocation, which makes the benchmarks worse but I think is acceptable in practice. 

In order to prove this I changed the benchmark script a bit to do static matching on a router without any versioned routes, and static matching on a separate router with versioned routes. The new benchmarks:


case | master | constraints
-- | -- | --
lookup static route | 41,082,348 ops/sec ±1.00% (586 runs sampled) | 45,136,281 ops/sec ±0.53% (592 runs sampled)
lookup dynamic route | 3,726,261 ops/sec ±1.03% (578 runs sampled) | 3,814,964 ops/sec ±0.28% (589 runs sampled)
lookup dynamic multi-parametric route | 1,834,927 ops/sec ±1.24% (583 runs sampled) | 2,000,081 ops/sec ±0.48% (592 runs sampled)
lookup dynamic multi-parametric route with regex | 1,314,972 ops/sec ±1.87% (563 runs sampled) | 1,533,650 ops/sec ±0.20% (591 runs sampled)
lookup long static route | 3,111,047 ops/sec ±1.07% (587 runs sampled) | 3,240,176 ops/sec ±0.44% (591 runs sampled)
lookup long dynamic route | 2,370,022 ops/sec ±0.73% (579 runs sampled) | 2,478,511 ops/sec ±0.28% (587 runs sampled)
lookup static route on constrained router | 17,162,042 ops/sec ±1.12% (576 runs sampled) | 16,966,232 ops/sec ±0.67% (585 runs sampled)
lookup static versioned route | 4,030,416 ops/sec ±0.38% (588 runs sampled) | 2,750,519 ops/sec ±0.42% (588 runs sampled)
lookup static constrained (version & host) route | 3,916,773 ops/sec ±0.82% (588 runs sampled) | 2,789,442 ops/sec ±0.22% (592 runs sampled)
find static route | 33,955,710 ops/sec ±0.99% (582 runs sampled) | 32,015,040 ops/sec ±0.84% (578 runs sampled)
find dynamic route | 4,253,495 ops/sec ±0.69% (585 runs sampled) | 4,317,023 ops/sec ±0.39% (588 runs sampled)
find dynamic multi-parametric route | 2,362,605 ops/sec ±0.60% (589 runs sampled) | 2,284,194 ops/sec ±0.45% (587 runs sampled)
find dynamic multi-parametric route with regex | 1,658,420 ops/sec ±0.54% (587 runs sampled) | 1,570,746 ops/sec ±0.25% (587 runs sampled)
find long static route | 4,435,708 ops/sec ±1.00% (583 runs sampled) | 4,327,720 ops/sec ±0.50% (591 runs sampled)
find long dynamic route | 3,331,895 ops/sec ±0.61% (586 runs sampled) | 3,132,098 ops/sec ±0.41% (584 runs sampled)
find long nested dynamic route | 1,419,802 ops/sec ±0.52% (591 runs sampled) | 1,328,724 ops/sec ±0.32% (589 runs sampled)
find long nested dynamic route with other method | 2,754,041 ops/sec ±0.68% (586 runs sampled) | 2,620,105 ops/sec ±0.54% (587 runs sampled)

How do y'all feel about the regression in performance for constrained routers in order to support more constraints?

index.js
airhorns: Updated to not make this call unless there are routes with constraints

mcollina (approved):
lgtm

This is semver-major.

airhorns:
@mcollina for the fastify side of this PR, I assume you don't want it to be a breaking change in the same way right? It is possible to keep it compatible I believe, just some work, and there will have to be new fastify API to allow passing the constraint strategies and route constraints as well.

mcollina:
Happy to land in a minor if it does not change any options. Maybe you can add `version` back to the route definition?

airhorns:
@delvedor would love to get your thoughts on this, happy to explain anything in detail if you are at all concerned!

airhorns:
@delvedor @mcollina anything I can do to help push this over the finish line? I've been running it in production for almost 6 weeks now and it's been working well. Happy to discuss if there's something missing as well!

airhorns:
@delvedor I know it's a busy season but is there something missing from this PR you need to see before it meets your standards? Would really love to get this shipped!

airhorns:
@delvedor @mcollina I would really love to ship this or some version of it! Is there anything else either high or low level you want to see addressed? If you want to do a different strategy of pluggable routers or something like that I'd be happy to discuss that too! @AyoubElk and I both worked pretty hard on this and would really love to see the community benefit!

mcollina:
We are waiting on @delvedor  :).

index.js
```diff
@@ -361,16 +349,16 @@ Router.prototype.off = function off (method, path) {
...
-  var handle = this.find(req.method, sanitizeUrl(req.url), this.versioning.deriveVersion(req, ctx))
+  var handle = this.find(req.method, sanitizeUrl(req.url), this.constrainer.deriveConstraints(req, ctx))
   if (handle === null) return this._defaultRoute(req, res, ctx)
   return ctx === undefined
     ? handle.handler(req, res, handle.params, handle.store)
     : handle.handler.call(ctx, req, res, handle.params, handle.store)
 }
 
-Router.prototype.find = function find (method, path, version) {
+Router.prototype.find = function find (method, path, derivedConstraints) {
```
delvedor, line 359: Why `derivedConstraints` must be defined?
Can't you assign it a default value?
```suggestion
Router.prototype.find = function find (method, path, derivedConstraints = {}) {
```
If we merge this change, the test should be updated as well.

delvedor (changes requested):
Hello! First of all, let me apologize for the massive delay in the review. You did a terrific job and I'm sorry to keep you waiting.
Overall the feature looks great! I've left a few comments while I take a second look at the compilation code.
From my benchmarks it looks like that the static route match has slowed down a bit (40M -> 38M), I'll re-run them.
The test coverage in `lib/strategies accept-host.js` can be improved, you can see the coverage report with `npm run test:report`.

index.js
AyoubElk: The `derivedConstraints` object is generated by `constrainer.deriveConstraints()` in `Router.lookup()`:
https://github.com/airhorns/find-my-way/blob/constraints/index.js#L352

And the `deriveConstraints` function is built [here](https://github.com/airhorns/find-my-way/blob/constraints/lib/constrainer.js#L74), and can return the value `undefined` if no constraint is in use.

delvedor: Sorry, not sure I follow, by looking at some test that you have updated, you can pass an empty object (and thus, use a default if the `derivedConstraints ` are not defined):
https://github.com/airhorns/find-my-way/blob/5344dd5fa2520016cbfb39dea1433f53a7a15cf5/test/issue-62.test.js#L15-L16

index.js
```diff
@@ -51,7 +51,24 @@ function Router (opts) {
...
+        mustMatchWhenDerived: true,
+        storage: opts.versioning.storage,
+        deriveConstraint: opts.versioning.deriveVersion,
+        validate (value) {
+          assert(typeof value === 'string', 'Version should be a string')
+        }
+      }
+    }
+  }
+  this.constrainer = new Constrainer(constraints)
```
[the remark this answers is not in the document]
airhorns: I'm fine with this approach, but it does mean a not-strictly-necessary breaking change for other users of this module that aren't fastify. Do we care about them?

test/constraint.custom-versioning.test.js
```diff
@@ -0,0 +1,63 @@
...
+  t.plan(2)
+
+  const findMyWay = FindMyWay({ constraints: { version: customVersioning } })
+
+  findMyWay.on('GET', '/', { constraints: { version: 'application/vnd.example.api+json;version=2' } }, (req, res, params) => {
+    t.ok(req.headers.accept, 'application/vnd.example.api+json;version=2')
+  })
+
+  findMyWay.on('GET', '/', { constraints: { version: 'application/vnd.example.api+json;version=3' } }, (req, res, params) => {
+    t.ok(req.headers.accept, 'application/vnd.example.api+json;version=3')
```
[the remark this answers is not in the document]
airhorns: Whoops, thanks!

index.js
airhorns: So this was deliberate for a couple of reasons. First, I wanted to make the API really clear in that you should almost definitely be using `lookup` and not `find` because what you pass in for this `derivedConstraints` object really matters. We can default it to something, but I don't think we really want to make this API ergonomic, because if you don't pass it, you are just turning off the constraint system without really knowing. If the contract of the router is to match constrained routes to requests, that feels kind of dangerous that it would just turn off that constraining part. Especially if we're going to make this a breaking change, I feel like it's a good opportunity to change the API to reflect the actual needs of the function. If that's weird I am happy to change it but it expressed intent to me which I think is important.

Second reason is just to avoid the extra statement and/or extra allocation, prioritizing performance over ergonomics.

I also kind of think `find` I think should be private to be honest, but I was trying to preserve some measure of backwards compatibility.

airhorns:
With respect to the benchmark regression, I thought that was just because I changed the structure of the benchmark just a little and thought it was apples to oranges, but I just compared the new benchmark structure on the master branch to this branch and noticed the same regression, I'll look into it. Womp womp.

index.js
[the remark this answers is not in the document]
delvedor: > Do we care about them?

We do! Reason why we are releasing this as a major and not a minor :)
In the changelog we'll write the upgrade path.

airhorns:
OK, added a few more microoptimizations that to my eye bring the benchmarks back in line. Just to note that a straight comparison against the master benchmarks isn't an apples to apples comparison, so I pushed a branch that is an apples to apples comparison here https://github.com/airhorns/find-my-way/tree/constraint-bench

```
inspector ~/C/find-my-way (constraints) ➜  git co constraint-bench
Switched to branch 'constraint-bench'
Your branch is up to date with 'origin/constraint-bench'.
inspector ~/C/find-my-way (constraint-bench) ➜  npm run bench

> find-my-way@3.0.5 bench
> node bench.js

lookup static route x 45,899,343 ops/sec ±0.43% (587 runs sampled)
lookup dynamic route x 3,934,461 ops/sec ±0.40% (592 runs sampled)
lookup dynamic multi-parametric route x 2,261,571 ops/sec ±0.26% (592 runs sampled)
lookup dynamic multi-parametric route with regex x 1,603,077 ops/sec ±0.85% (576 runs sampled)
lookup long static route x 3,572,855 ops/sec ±0.46% (583 runs sampled)
lookup long dynamic route x 2,529,801 ops/sec ±0.53% (583 runs sampled)
find static route x 37,720,045 ops/sec ±0.62% (574 runs sampled)
find dynamic route x 4,571,864 ops/sec ±0.52% (587 runs sampled)
find dynamic multi-parametric route x 2,489,602 ops/sec ±0.35% (592 runs sampled)
find dynamic multi-parametric route with regex x 1,820,313 ops/sec ±0.29% (593 runs sampled)
find long static route x 5,313,042 ops/sec ±0.29% (589 runs sampled)
find long dynamic route x 3,666,330 ops/sec ±0.63% (587 runs sampled)
find long nested dynamic route x 1,487,051 ops/sec ±0.56% (590 runs sampled)
find long nested dynamic route with other method x 2,940,246 ops/sec ±0.29% (587 runs sampled)
find long nested dynamic route x 1,483,649 ops/sec ±0.30% (589 runs sampled)
find long nested dynamic route with other method x 2,963,341 ops/sec ±0.31% (588 runs sampled)
inspector ~/C/find-my-way (constraint-bench) ➜  git co constraints
Switched to branch 'constraints'
Your branch is up to date with 'origin/constraints'.
inspector ~/C/find-my-way (constraints) ➜  npm run bench

> find-my-way@3.0.5 bench
> node bench.js

lookup static route x 46,104,542 ops/sec ±0.46% (590 runs sampled)
lookup dynamic route x 3,695,481 ops/sec ±0.21% (591 runs sampled)
lookup dynamic multi-parametric route x 2,054,533 ops/sec ±0.19% (590 runs sampled)
lookup dynamic multi-parametric route with regex x 1,511,028 ops/sec ±0.22% (588 runs sampled)
lookup long static route x 3,388,233 ops/sec ±0.35% (590 runs sampled)
lookup long dynamic route x 2,507,811 ops/sec ±0.31% (586 runs sampled)
lookup static route on constrained router x 18,078,276 ops/sec ±0.45% (585 runs sampled)
lookup static versioned route x 2,789,047 ops/sec ±0.36% (587 runs sampled)
lookup static constrained (version & host) route x 2,828,369 ops/sec ±0.29% (588 runs sampled)
find static route x 37,346,082 ops/sec ±0.69% (577 runs sampled)
find dynamic route x 4,293,511 ops/sec ±0.32% (589 runs sampled)
find dynamic multi-parametric route x 2,360,816 ops/sec ±0.29% (588 runs sampled)
find dynamic multi-parametric route with regex x 1,669,932 ops/sec ±0.29% (589 runs sampled)
find long static route x 4,772,811 ops/sec ±0.43% (590 runs sampled)
find long dynamic route x 3,438,186 ops/sec ±0.22% (593 runs sampled)
find long nested dynamic route x 1,454,251 ops/sec ±0.20% (593 runs sampled)
find long nested dynamic route with other method x 2,908,903 ops/sec ±0.24% (593 runs sampled)
find long nested dynamic route x 1,462,207 ops/sec ±0.18% (593 runs sampled)
find long nested dynamic route with other method x 2,911,268 ops/sec ±0.22% (594 runs sampled)
```

 I would echo what other contributors have noticed which is that the benchmarks seem to be quite a bit more variable than the error margins inside them would otherwise report. I would also note that the `lookup` benchmarks are faster than the `find` benchmarks, which does not make much sense because `lookup` does more work. I suspect that the JIT has something to do with this, but it suggests that this library might need a different benchmarking strategy.

index.js
delvedor: I would like to keep the `find` invocation free of additional params when constraints are not needed. There are use cases where find is more desirable compared to lookup.

airhorns: > I would like to keep the find invocation free of additional params when constraints are not needed. There are use cases where find is more desirable compared to lookup.

That means that the caller of `find` needs to be aware of if the router has constraints or not, which I think is kind of an anti-pattern. If someone is using `find-my-way` as part of a small system where they definitely know no routes are constrained, then what you're suggesting would make that a teensy bit easier, but is that really worth optimizing for? Anyone using `find-my-way` as part of a bigger ecosystem won't know what routes are registered ahead of time, in which case the API asking for constraint values is actually a positive as it forces them to implement the interface `find-my-way` needs to do it's job. 

Defaulting it also adds an extra object allocation which genuinely affects performance significantly, so we could do that if you're ok accepting the performance regression.

airhorns: > I'm lost. So passing findMyWay.find('GET', '/a/b') should work and disable all possible constraints?

It will work, correct, because it gets interpreted as `findMyWay.find('GET', '/a/b', undefined)`, which is what `.lookup` would pass when there are no constrained routes in use. The `undefined` there is what the constrainer produces when there aren't any constrained routes to avoid allocating an object or fishing out the version and host from the request. We needed to heavily optimize how the constrainer works to prevent virtually any extra work being done to support constraints when there are no routes to meet your folks' stringent performance requirements. Totally fine doing that, but we kind of had to which is why it is a bit of a strange API. 

> Constraints are niche use case and I think requiring this additional parameter for this is probably not needed. If undefined is a valid value, let's add a test to cover it and that'd be ok.

If you build a thing that uses find-my-way and doesn't pass a value for constraints, it is a bug in waiting though, isn't that ... bad? I think find-my-way really lends itself to being used as part of other resuable systems (like fastify, restify, etc), where the author of the reusable system can't necessarily guarantee what a different author might add as routes. @mcollina writes fastify, @airhorns writes routes, so if @airhorns decides to use constraints and @mcollina's fastify implementation doesn't pass the right stuff to find-my-way, the contract is broken. Why would we want to open up the API contract to that? 

I'll add the undefined test case, I think that's a great idea, but I think it'd be better to represent the 3rd parameter as a real thing that really should be passed if you're using `.find` in the docs and in the tests for ^ that reason.

Also, this should probably be a larger conversation and doesn't need to be solved now, but this problem would also go away if we privatized `.find` or changed it's API to be more useful. I think if you want a public`.find`, it should maybe take a request, not a set of constraints, and derive the constraints itself, as that would make this whole discussion moot. I am not really sure what the use cases for finding routes outside the context of a request is, especially if constraints are a thing, because you'd have to derive them yourself.

README.md
```diff
@@ -110,21 +178,21 @@ const customVersioning = {
...
 }
 
-const router = FindMyWay({ versioning: customVersioning });
+const router = FindMyWay({ constraints: { version: customVersioning } });
 ```
 
 The custom strategy object should contain next properties:
 * `storage` - the factory function for the Storage of the handlers based on their version.
-* `deriveVersion` - the function to determine the version based on the request
+* `deriveConstraint` - the function to determine the version based on the request
```
mcollina, line 192: I think "version" is not the right term here anymore.

index.js
mcollina: I think we should just document:

1. that the 3rd argument is optional, i.e. `findMyWay.find('GET', '/a/b')` works.
2. using `findMyWay.find('GET', '/a/b', undefined)` would lead to _unspecified_ behavior if there are any constraints in the system.

airhorns: Ok, I'll do that!

mcollina (approved):
lgtm

airhorns:
ping @delvedor care to take another look?

delvedor (approved):
LGTM

Before merging this I would kindly ask you to remove the not necessary empty object you have added in some test when calling `find`, given that it's not required we should not have it there, otherwise we risk to create a false positive test :)

airhorns:
> LGTM

Woo!

> Before merging this I would kindly ask you to remove the not necessary empty object you have added in some test when calling find, given that it's not required we should not have it there, otherwise we risk to create a false positive test :)

I think I might have been unclear about what that value is actually for because I consider that arg very much required. For the system to meet its contract, the caller needs to think about what is passed for that argument and pass something, it just be that the value passed is undefined for performance reasons. Most direct callers of `.find` aren't going to be finding one hardcoded handler, it's going to be some higher level function that uses find my way to route to any of the routes registered, which means the value for the constraints argument should probably be dynamic, not just hardcoded to undefined or omitted. That's why I went through and added it to every test. 

The reason is that within the realm of this library, I think constraints must be a thing that contributors care about. If they go to change the constraints implementation, or change something else that passes code to and from the bits that care about constraints, they should have to think about how the constraints are going to be passed in, and how to keep the performance of the constraint derivation high for whatever new thing they want to do. I think making it easy to ignore constraints is the wrong thing to do in this library, because (once merged) this library has committed to supporting them. I think if users of find-my-way want to not care about constraints, that's fine, they can just always pass undefined, but that's different than how we test and communicate to new contributors.

That said, I don't want to be a stick in the mud, so if it's really important to you that the tests don't get updated I can go through and revert those changes.

delvedor:
> That said, I don't want to be a stick in the mud, so if it's really important to you that the tests don't get updated I can go through and revert those changes.

Yes, it's important. If both behaviors are supported, then the test should verify that.

airhorns:
I agree its important, and there's tests that test the behaviour with and without the parameter. If that's all thats required we're good to merge, but are you saying you want all the test cases to implicitly test the undefined value, and then have only some tests that test with the value?

airhorns:
ping @delvedor is the final decision to remove the parameter from all the tests or can we merge?

delvedor:
Sorry to be a PITA, but if this pr is not changing the default behavior, those tests **should not** be updated.

airhorns:
Again -- it isn't that the default behaviour hasn't change, it has. Callers of `.find` should and will need to think about what to pass for the third parameter. Every call to `lookup` will call `find` with a value for the third parameter. It is required in the human sense of the word, just not in the JS VM sense of the word. I get that test cleanness is a thing though -- updated to remove the changes to `.find`! Any other blockers?

mcollina (approved):
lgtm

airhorns:
ping @delvedor anything else you'd like to see before :shipit: ? Would really like to get this in!

delvedor (approved):
LGTM 🚀

airhorns:
@AyoubElk would you mind giving one last review? Then I think we're ready to merge!

README.md
```diff
@@ -96,36 +96,105 @@ const router = require('find-my-way')({
...
-  // storage factory
+  // replace the built in version strategy
+  name: 'version',
+  // provide a storage factory to store handlers in a simple way
   storage: function () {
     let versions = {}
     return {
       get: (version) => { return versions[version] || null },
-      set: (version, store) => { versions[version] = store },
+      set: (version, handler) => { versions[version] = handler },
```
AyoubElk, line 177: Nitpicking, but I think `store` is a more descriptive name compared to `handler` since we actually set a bitmap of handler indexes here.

```suggestion
      set: (version, store) => { versions[version] = store },
```

AyoubElk (changes requested):

AyoubElk:
LGTM

airhorns:
Ok great -- anything left before merging?

airhorns:
ping @delvedor can you hit the merge button? Or if there is anything outstanding let me know!

delvedor:
@airhorns there are also others pr that needs to be merged and since this one is a breaking change I'm wondering about if we should do a minor first and then cut the major or directly go all in with the major.
What do you think @mcollina?

mcollina:
I would just ship a major and then land the major update in FAstify.

PR #174, opened 2020-12-15
feature: add setDefaultRoute method

AyoubElk:
This change allows updating the `defaultRoute` method after the router instance is created.
It is needed [here](https://github.com/fastify/fastify/pull/2733) (relevant part of the discussion [here](https://github.com/fastify/fastify/pull/2733#pullrequestreview-550145862))

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,12 +33,8 @@ function Router (opts) {
   }
   opts = opts || {}
 
-  if (opts.defaultRoute) {
-    assert(typeof opts.defaultRoute === 'function', 'The default route must be a function')
-    this.defaultRoute = opts.defaultRoute
-  } else {
-    this.defaultRoute = null
-  }
+  this.defaultRoute = null
+  this.setDefaultRoute(opts.defaultRoute)
 
   if (opts.onBadUrl) {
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -588,6 +584,13 @@ Router.prototype._defaultRoute = function (req, res, ctx) {
   }
 }
 
+Router.prototype.setDefaultRoute = function (defaultRoute) {
+  if (defaultRoute) {
+    assert(typeof defaultRoute === 'function', 'The default route must be a function')
+    this.defaultRoute = defaultRoute
+  }
+}
```
mcollina, line 592: I would implement it with a `defineProperty` with a getter/setter instead. Note that this might not be needed at all as somebody could already do `router.defaultRoute = fn` - so maybe this PR is not needed?

AyoubElk: True, it's possible to update it directly inside `fastify` and expose the methods needed by `fastify-websocket` there.
I thought it would be more correct to do it here to keep its validation (assert) encapsulated here without repeating it inside `fastify`, but yeah this doesn't seem necessary after all.

mcollina: I would just update `router.defaultRoute` in fastify.

AyoubElk: Got it, thanks

PR #176, opened 2021-01-13
Add double colon handling (support for paths with colons)

morigs:
Resolves delvedor#175


mcollina (commented):
Could you run the benchmark before and after the change and post the results?

mcollina:
Good work!

morigs:
Sure. My results are the following

**Before**
```
lookup static route x 46,229,112 ops/sec ±0.54% (591 runs sampled)
lookup dynamic route x 2,886,813 ops/sec ±0.43% (586 runs sampled)
lookup dynamic multi-parametric route x 1,614,329 ops/sec ±0.18% (589 runs sampled)
lookup dynamic multi-parametric route with regex x 1,115,574 ops/sec ±0.18% (587 runs sampled)
lookup long static route x 2,904,037 ops/sec ±0.20% (588 runs sampled)
lookup long dynamic route x 2,018,473 ops/sec ±0.19% (590 runs sampled)
lookup static versioned route x 6,800,830 ops/sec ±0.19% (586 runs sampled)
find static route x 30,635,586 ops/sec ±1.11% (573 runs sampled)
find dynamic route x 3,362,088 ops/sec ±0.15% (589 runs sampled)
find dynamic multi-parametric route x 1,899,912 ops/sec ±0.20% (589 runs sampled)
find dynamic multi-parametric route with regex x 1,245,699 ops/sec ±0.19% (590 runs sampled)
find long static route x 4,195,288 ops/sec ±0.20% (589 runs sampled)
find long dynamic route x 2,691,891 ops/sec ±0.18% (590 runs sampled)
find static versioned route x 8,167,230 ops/sec ±0.22% (583 runs sampled)
find long nested dynamic route x 1,043,135 ops/sec ±0.13% (590 runs sampled)
find long nested dynamic route with other method x 2,010,740 ops/sec ±0.16% (589 runs sampled)
```

**After**
```
lookup static route x 46,398,290 ops/sec ±0.53% (590 runs sampled)
lookup dynamic route x 2,882,147 ops/sec ±0.18% (589 runs sampled)
lookup dynamic multi-parametric route x 1,575,300 ops/sec ±0.17% (591 runs sampled)
lookup dynamic multi-parametric route with regex x 1,182,742 ops/sec ±0.16% (590 runs sampled)
lookup long static route x 2,946,382 ops/sec ±0.32% (588 runs sampled)
lookup long dynamic route x 2,027,316 ops/sec ±0.32% (589 runs sampled)
lookup static versioned route x 6,859,823 ops/sec ±0.21% (586 runs sampled)
find static route x 32,015,062 ops/sec ±0.59% (576 runs sampled)
find dynamic route x 3,404,869 ops/sec ±0.16% (590 runs sampled)
find dynamic multi-parametric route x 1,855,533 ops/sec ±0.14% (590 runs sampled)
find dynamic multi-parametric route with regex x 1,310,752 ops/sec ±0.15% (591 runs sampled)
find long static route x 4,085,360 ops/sec ±0.14% (590 runs sampled)
find long dynamic route x 2,673,217 ops/sec ±0.15% (591 runs sampled)
find static versioned route x 8,030,499 ops/sec ±0.20% (581 runs sampled)
find long nested dynamic route x 1,034,793 ops/sec ±0.18% (592 runs sampled)
find long nested dynamic route with other method x 1,981,866 ops/sec ±0.13% (592 runs sampled)
```

index.js
```diff
@@ -119,6 +119,13 @@ Router.prototype._on = function _on (method, path, opts, handler, store) {
     // search for parametric or wildcard routes
     // parametric route
     if (path.charCodeAt(i) === 58) {
+      if (i !== len - 1 && path.charCodeAt(i + 1) === 58) {
```
delvedor, line 122: I'm wondering if we can check if the preceding char is a slash (47) and if not avoid creating the parametric node.

eg:
```
/foo/:bar => char before : is / => parametric route
/foo/bar:baz => char before : is not / => static route
```

morigs: I was thinking about that. But what if I want a parameter following some static string?
Maybe something like `/tests/issue-:num(^\d+$)/description` to route to `/tests/issue-175/description`

delvedor: The same thing, we can check for the preceding char code (45 in this case) and act accordingly.
I feel that having two columns is counterintuitive and obscure. I prefer to have a transparent API and the doc explains how to achieve the specific use cases like the one you've mentioined.

morigs: So which syntax do you think is more intuitive?
I agree that two colons is not very intuitive. 
But I was thinking a lot of people came to fastify from express.js and Koa. (https://github.com/expressjs/express/issues/3857)
I could replace two colons with backslash for example

delvedor: I think the more intuitive is the one I've described in my first comment.
I agree that familiarity with express/koa is nice to have, but not when we disagree with a pattern.
@mcollina what do you think?

mcollina: A change that would not require an additional `charCodeAt()` in the hot path would be better. However, this is not causing any regression so maybe this is also ok.

morigs: @delvedor according to the current docs it's possible to have any number of parameters within the same couple of slash. And there are no exact delimiters (slash or something).
Example from the readme:
`router.on('GET', '/example/at/:hour(^\\d{2})h:minute(^\\d{2})m', (req, res, params) => {}))`
In this example the char before `:` is `h`. So based on the structure of the route, we cannot know for sure if I want to declare a new parameter or just use a colon.

mcollina (approved):
lgtm, this is ok as-is

index.js
delvedor: Great point @morigs!

delvedor (approved):
LGTM

Issue #179, opened 2021-02-11
Question: Object as value in custom constraint 


matthyk:
Since version 4.0.0, it is possible to write your own constraint strategies. In the index.d.ts, however, it is specified that the values for the constraints must be strings. So my question is whether it is safe to also take object as values, as in the example below.

Current typing in index.d.ts:

```javascript
interface ConstraintStrategy<V extends HTTPVersion> {
    name: string,
    mustMatchWhenDerived?: boolean,
    storage() : {
      get(version: String) : Handler<V> | null,
      set(version: String, store: Handler<V>) : void,
      del(version: String) : void,
      empty() : void
    },
    validate(value: unknown): void,
    deriveConstraint<Context>(req: Req<V>, ctx?: Context) : String,
  }
```
Example with an object as key instead of a string:

```javascript

interface MyObject {
    myProperty: string;
    mySecondProperty: string;
    myNumberProperty: number;
}

interface ConstraintStrategy<V extends HTTPVersion> {
    name: string,
    mustMatchWhenDerived?: boolean,
    storage() : {
      get(version: MyObject) : Handler<V> | null,
      set(version: MyObject, store: Handler<V>) : void,
      del(version: MyObject) : void,
      empty() : void
    },
    validate(value: unknown): void,
    deriveConstraint<Context>(req: Req<V>, ctx?: Context) : MyObject,
  }
```

Could be typed in the index.d.ts like this:

```javascript
interface ConstraintStrategy<V extends HTTPVersion, T = string> {
    name: string,
    mustMatchWhenDerived?: boolean,
    storage() : {
      get(version: T) : Handler<V> | null,
      set(version: T, store: Handler<V>) : void,
      del(version: T) : void,
      empty() : void
    },
    validate(value: unknown): void,
    deriveConstraint<Context>(req: Req<V>, ctx?: Context) : T,
  }
```

mcollina:
I would say that the best way is to send a PR and figure things out :).

matthyk:
Created PR #180

PR #180, opened 2021-02-12
update ConstraintStrategy typing

matthyk:
In #179 it was asked whether it is possible to use objects as values in custom constraints. After looking through the code, this should not be a problem. Only the typings have to be updated.
The default value of the generic could be `string` or `any`. 

`any` would match the typing of the RouteOption. 

```javascript
interface RouteOptions {
    constraints?: { [key: string]: any }
  }
```

But `string` would provide a backwards compatibility.


mcollina (changes requested):
Please update the tests as well.

matthyk:
Should I only update the tsd tests or also add a test to test custom constraints with objects as values?

matthyk:
Which tests exactly need to be updated? For me it looks like it would make most sense to add a new tests for custom constraints with objects. ( like `constraint.custom-with-objects.test` )

mcollina:
Adding new tests is totally fine

mcollina:
We just need to test the types: https://github.com/delvedor/find-my-way/blob/master/test/types/router.test-d.ts

mcollina (approved):
lgtm

index.d.ts
```diff
@@ -57,17 +57,17 @@ declare namespace Router {
...
 
-  interface ConstraintStrategy<V extends HTTPVersion> {
+  interface ConstraintStrategy<V extends HTTPVersion, T = string> {
     name: string,
     mustMatchWhenDerived?: boolean,
     storage() : {
-      get(version: String) : Handler<V> | null,
-      set(version: String, store: Handler<V>) : void,
-      del(version: String) : void,
+      get(version: T) : Handler<V> | null,
```
airhorns, line 64: Would you mind updating this variable to be named `value` or something less version specific? Not strictly a result of your change but it is a bit more confusing now that the type of this piece is generic

airhorns (approved):
LGTM, nice change!

index.d.ts
matthyk: Perhaps the name of the `store` variable should also be changed, as this has simply been taken over from the semver-store?

airhorns: Oh -- yep definitely, that's a `handler` not a `store`. Thanks for leaving it better than you found it!!

delvedor (approved):
LGTM

PR #182, opened 2021-03-09
Pretty print bugfix and revision

polymathca:
I found some unexpected behaviour in the way routes were displayed while I was testing this feature addition for `fastify-autoload` (https://github.com/fastify/fastify-autoload/pull/134)

The route parameters weren't being properly placed when drawing the tree if there was no root item. An example can be found in the PR linked above. I added a test for a specific instance where I was able to get it to break.

I have also added an additional tree-drawing function which uses the `findMyWay.routes` array instead of the radix tree, as it might be more intuitive to some users because it doesn't break trees by common prefixes (e.g. `home, homer, honour` in place of `ho->me, ho->me->r, ho->nour`)

If the added function is appropriate, I will add some tests for it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,7 +409,7 @@ router.find('GET', '/example', { host: 'fastify.io', version: '1.x' })
 ```
 
 <a name="pretty-print"></a>
-#### prettyPrint([{ commonPrefix: true }])
+#### prettyPrint([{ commonPrefix: false }])
 Prints the representation of the internal radix tree, useful for debugging.
 ```js
 findMyWay.on('GET', '/test', () => {})
@@ -430,7 +430,7 @@ console.log(findMyWay.prettyPrint())
 `prettyPrint` accepts an optional setting to use the internal routes array to render the tree.
 
 ```js
-console.log(findMyWay.prettyPrint({ commonPrefix: true }))
+console.log(findMyWay.prettyPrint({ commonPrefix: false }))
 // └── / (-)
 //     ├── test (GET)
 //     │   └── /hello (GET)
--- a/README.md
+++ b/README.md
@@ -409,7 +409,7 @@ router.find('GET', '/example', { host: 'fastify.io', version: '1.x' })
 ```
 
 <a name="pretty-print"></a>
-#### prettyPrint([{ routesArray: true }])
+#### prettyPrint([{ commonPrefix: true }])
 Prints the representation of the internal radix tree, useful for debugging.
 ```js
 findMyWay.on('GET', '/test', () => {})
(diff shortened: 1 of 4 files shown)


lib/pretty-print.js
```diff
@@ -1,5 +1,118 @@
+/* eslint-disable no-multi-spaces */
```
mcollina, line 1: Can you please add `'use strict'` at the top?

mcollina (commented):
Good work, I've just left a nit!

mcollina:
I'm ok with the added function.

lib/pretty-print.js
polymathca: Done!

polymathca:
Two more notes:

I had to modify the constrained parametric route test, as it wasn't including the preceding slashes in the expected output on the constrained paths. I don't think that should be problematic, but it's worth mentioning. Link to that change here:

https://github.com/polymathca/find-my-way/blob/4a92e9af8fc6f0e091a75aa1ef0e3e0ef886db9a/test/pretty-print.test.js#L124-L129

The `routesArray: true` output explicitly indicates nodes without methods attached with a `(-)` which I feel is a more explicit way to show no handler is attached at that level. I can revise that if it's not correct (I left that as-is in the original function).

README.md
```diff
@@ -409,18 +409,34 @@ router.find('GET', '/example', { host: 'fastify.io', version: '1.x' })
 ```
 
 <a name="pretty-print"></a>
-#### prettyPrint()
+#### prettyPrint([{ routesArray: true }])
```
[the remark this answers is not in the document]
polymathca: Good call!

mcollina (approved):
lgtm

delvedor (changes requested):
Great work! Can you flip the way the option works?
`commonPrefix: false` shows the full name of each path chunk
`commonPrefix: true` shows how the prefix is split internally (which is the default)

polymathca:
> 
> 
> Great work! Can you flip the way the option works?
> `commonPrefix: false` shows the full name of each path chunk
> `commonPrefix: true` shows how the prefix is split internally (which is the default)

For sure! Or do you think it would be better to rename the option (e.g. `useFullPaths`) so we aren't altering the current default behaviour of showing the radix tree when the function is called without options?

delvedor:
`commonPrefix` is fine, the default behavior is `commonPrefix:true`

polymathca:
Sounds good, I added a test that explicitly shows the behaviour of `commonPrefix: true` and `commonPrefix: false`, just for clarity

mcollina (approved):
lgtm

polymathca:
I tidied up the branch-drawing somewhat so the `commonPrefix: true` output doesn't insert spaces in new items with constraints, and the `commonPrefix: false` version now outputs constrained routes on separate lines for clarity.

mcollina (approved):
lgtm

delvedor (approved):
LGTM

PR #183, opened 2021-03-13
Widen type defs to allow for nonstring version constraints

danielmoore:
find-my-way's design makes it agnostic as to the specific type and shape of a version, however the bundled types constrain versions to a string. This PR does three things:

1. Provides a test showing that non-string versions are supported.
1. Widens the typings to a generic, so that implemented strategies are consistently typed, with a string version as a default.
1. Marks `validate` as optional, because it is.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -10,7 +10,7 @@ declare namespace Router {
     V1 = 'http1',
     V2 = 'http2'
   }
-
+  
   type HTTPMethod =
     | 'ACL'
     | 'BIND'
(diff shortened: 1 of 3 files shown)


mcollina:
can you add a test for the type in test/types/router.test-d.ts?

mcollina:
good work!

danielmoore:
Thanks! I've added some type tests. I'm new to type-testing, so any feedback you have on it would be great.

mcollina (approved):
lgtm

index.d.ts
```diff
@@ -57,17 +57,17 @@ declare namespace Router {
     store: any
   ) => void;
 
-  interface ConstraintStrategy<V extends HTTPVersion> {
+  interface ConstraintStrategy<V extends HTTPVersion, ConstraintVersion = string> {
```
airhorns, line 60: I think it might make more sense to call this property `ConstraintValue` or something like that. I know it says `version` in all the types below, but these strategies can actually be used for other constraints than version, like `host` or maybe the `Accept` header someday. I think the name of that argument in the types should be changed too.

index.d.ts
```diff
@@ -91,7 +91,7 @@ declare namespace Router {
     ): void;
 
     constraints? : {
-      [key: string]: ConstraintStrategy<V>
+      [key: string]: ConstraintStrategy<V, unknown>
```
airhorns, line 94: This seems like it might be a breaking change, do the fastify type tests still pass with this in place?

ivan-tymoshenko:
I guess it was implemented here https://github.com/delvedor/find-my-way/pull/180. We can close it.

PR #185, opened 2021-04-12
Feature: adding support to display fastify hooks

polymathca:
Re: https://github.com/fastify/fastify/issues/2993

Adds ability to display hooks in `prettyPrint` output.

```
commonPrefix: false

   ├── /pages (GET)
   │   • (onTimeout) ["timeoutHook"]
   │   • (onRequest) ["anonymous","authenticate_token"]
   │   ├── / (GET)
   │   │   • (onTimeout) ["timeoutHook"]
   │   │   • (onRequest) ["anonymous","authenticate_token"]
   │   ├── /:id (GET)
   │   │   • (onTimeout) ["timeoutHook"]
   │   │   • (onRequest) ["anonymous","authenticate_token"]
   │   │   └── / (GET)
   │   │       • (onTimeout) ["timeoutHook"]
   │   │       • (onRequest) ["anonymous","authenticate_token"]
   │   ├── /new (POST)
   │   │   • (onTimeout) ["timeoutHook"]
   │   │   • (onRequest) ["anonymous","authenticate_token"]
   │   ├── /testing (GET) {"version":"1.0"}
   │   │   • (onTimeout) ["timeoutHook"]
   │   │   • (onRequest) ["anonymous","authenticate_token"]
   │   │   /testing (GET, PUT) {"version":"1.1"}
   │   │   • (onTimeout) ["timeoutHook"]
   │   │   • (onRequest) ["anonymous","authenticate_token"]
   │   └── /update (POST)
   │       • (onTimeout) ["timeoutHook"]
   │       • (onRequest) ["anonymous","authenticate_token"]
```

```
commonPrefix: true

   │   └── p
   │       ├── ages (GET)
   │       │   • (onTimeout) ["timeoutHook"]
   │       │   • (onRequest) ["anonymous","authenticate_token"]
   │       │   └── / (GET)
   │       │       • (onTimeout) ["timeoutHook"]
   │       │       • (onRequest) ["anonymous","authenticate_token"]
   │       │       ├── testing (GET) {"version":"1.0"}
   │       │       │   • (onTimeout) ["timeoutHook"]
   │       │       │   • (onRequest) ["anonymous","authenticate_token"]
   │       │       │   testing (GET) {"version":"1.1"}
   │       │       │   • (onTimeout) ["timeoutHook"]
   │       │       │   • (onRequest) ["anonymous","authenticate_token"]
   │       │       ├── :id (GET)
   │       │       │   • (onTimeout) ["timeoutHook"]
   │       │       │   • (onRequest) ["anonymous","authenticate_token"]
   │       │       │   └── / (GET)
   │       │       │       • (onTimeout) ["timeoutHook"]
   │       │       │       • (onRequest) ["anonymous","authenticate_token"]
   │       │       ├── new (POST)
   │       │       │   • (onTimeout) ["timeoutHook"]
   │       │       │   • (onRequest) ["anonymous","authenticate_token"]
   │       │       └── update (POST)
   │       │           • (onTimeout) ["timeoutHook"]
   │       │           • (onRequest) ["anonymous","authenticate_token"]
```
Enabled by `opts.includeHooks` and specified by `opts.hooksArray` (the latter defaults to `fastify/Hooks.supportedHooks`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ You can assign a `buildPrettyPrint` function to sanitize a route's `store` objec
 ```js
 
 const privateKey = new Symbol('private key')
-const store = { token: '12345' }
+const store = { token: '12345', [privateKey]: 'private value' }
 
 const router = require('find-my-way')({
   buildPrettyMeta: route => {
@@ -112,7 +112,7 @@ const router = require('find-my-way')({
       if (typeof k === 'symbol') delete cleanMeta[k]
     })
 
-    return cleanMeta
+    return cleanMeta // this will show up in the pretty print output!
   }
 })
 
--- a/README.md
+++ b/README.md
@@ -96,6 +96,37 @@ const router = require('find-my-way')({
 })
 ```
 
+You can assign a `buildPrettyPrint` function to sanitize a route's `store` object to use with the `prettyPrint` functions. This function should accept a single object and return an object.
+
+```js
+
+const privateKey = new Symbol('private key')
+const store = { token: '12345' }
+
+const router = require('find-my-way')({
+  buildPrettyMeta: route => {
+    const cleanMeta = Object.assign({}, route.store)
+
+    // remove private properties
+    Object.keys(cleanMeta).forEach(k => {
+      if (typeof k === 'symbol') delete cleanMeta[k]
+    })
+
+    return cleanMeta
+  }
+})
+
+store[privateKey] = 'private value'
+router.on('GET', '/hello_world', (req, res) => {}, store)
+
+router.prettyPrintRouteArray()
+
+//└── / (-)
+//    └── hello_world (GET)
+//        • (token) "12345"
+
+```
+
 ## Constraints
 
 `find-my-way` supports restricting handlers to only match certain requests for the same path. This can be used to support different versions of the same route that conform to a [semver](#semver) based versioning strategy, or restricting some routes to only be available on hosts. `find-my-way` has the semver based versioning strategy and a regex based hostname constraint strategy built in.
@@ -444,7 +475,9 @@ console.log(findMyWay.prettyPrint({ commonPrefix: false }))
 To include a display of the `store` data passed to individual routes, the 
 option `includeMeta` may be passed. If set to `true` all items will be
 displayed, this can also be set to an array specifying which keys (if
-present) to be displayed.
+present) should be displayed. This information can be further sanitized
+by specifying a `buildPrettyMeta` function which consumes and returns
+an object.
 
 ```js
 findMyWay.on('GET', '/test', () => {}, { onRequest: () => {}, authIDs => [1,2,3] })
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ const router = require('find-my-way')({
 })
 ```
 
-You can assign a `buildPrettyPrint` function to sanitize a route's `store` object to use with the `prettyPrint` functions. This function should accept a single object and return an object.
+You can assign a `buildPrettyMeta` function to sanitize a route's `store` object to use with the `prettyPrint` functions. This function should accept a single object and return an object.
 
 ```js
 
--- a/README.md
+++ b/README.md
@@ -409,8 +409,9 @@ router.find('GET', '/example', { host: 'fastify.io', version: '1.x' })
 ```
 
 <a name="pretty-print"></a>
-#### prettyPrint([{ commonPrefix: false }])
+#### prettyPrint([{ commonPrefix: false, includeMeta: true || [] }])
 Prints the representation of the internal radix tree, useful for debugging.
+
 ```js
 findMyWay.on('GET', '/test', () => {})
 findMyWay.on('GET', '/test/hello', () => {})
@@ -427,13 +428,47 @@ console.log(findMyWay.prettyPrint())
 //     └── update (PUT)
 ```
 
-`prettyPrint` accepts an optional setting to use the internal routes array to render the tree.
+`prettyPrint` accepts an optional setting to use the internal routes array
+to render the tree.
 
 ```js
 console.log(findMyWay.prettyPrint({ commonPrefix: false }))
 // └── / (-)
 //     ├── test (GET)
 //     │   └── /hello (GET)
+//     ├── testing (GET)
+//     │   └── /:param (GET)
+//     └── update (PUT)
+```
+
+To include a display of the `store` data passed to individual routes, the 
+option `includeMeta` may be passed. If set to `true` all items will be
+displayed, this can also be set to an array specifying which keys (if
+present) to be displayed.
+
+```js
+findMyWay.on('GET', '/test', () => {}, { onRequest: () => {}, authIDs => [1,2,3] })
+findMyWay.on('GET', '/test/hello', () => {}, { token: 'df123-4567' })
+findMyWay.on('GET', '/testing', () => {})
+findMyWay.on('GET', '/testing/:param', () => {})
+findMyWay.on('PUT', '/update', () => {})
+
+console.log(findMyWay.prettyPrint({ commonPrefix: false, includeMeta: ['onRequest'] }))
+// └── /
+//     ├── test (GET)
+//     │   • (onRequest) "anonymous()"
+//     │   ├── /hello (GET)
+//     │   └── ing (GET)
+//     │       └── /:param (GET)
+//     └── update (PUT)
+
+console.log(findMyWay.prettyPrint({ commonPrefix: true, includeMeta: true }))
+// └── / (-)
+//     ├── test (GET)
+//     │   • (onRequest) "anonymous()"
+//     │   • (authIDs) [1,2,3]
+//     │   └── /hello (GET)
+//     │       • (token) "df123-4567"
 //     ├── testing (GET)
 //     │   └── /:param (GET)
 //     └── update (PUT)
--- a/index.js
+++ b/index.js
@@ -580,7 +580,7 @@ Router.prototype._onBadUrl = function (path) {
 
 Router.prototype.prettyPrint = function (opts = {}) {
   opts.commonPrefix = opts.commonPrefix === undefined ? true : opts.commonPrefix // default to original behaviour
-  if (!opts.commonPrefix) return prettyPrintRoutesArray(this.routes, opts, this)
+  if (!opts.commonPrefix) return prettyPrintRoutesArray.call(this, this.routes, opts)
   const root = {
     prefix: '/',
     nodes: [],
@@ -595,7 +595,7 @@ Router.prototype.prettyPrint = function (opts = {}) {
 
   compressFlattenedNode(root)
 
-  return prettyPrintFlattenedNode(root, '', true, opts, this)
+  return prettyPrintFlattenedNode.call(this, root, '', true, opts)
 }
 
 for (var i in http.METHODS) {
--- a/index.js
+++ b/index.js
@@ -573,7 +573,7 @@ Router.prototype._onBadUrl = function (path) {
 
 Router.prototype.prettyPrint = function (opts = {}) {
   opts.commonPrefix = opts.commonPrefix === undefined ? true : opts.commonPrefix // default to original behaviour
-  if (!opts.commonPrefix) return prettyPrintRoutesArray(this.routes)
+  if (!opts.commonPrefix) return prettyPrintRoutesArray(this.routes, opts)
   const root = {
     prefix: '/',
     nodes: [],
--- a/index.js
+++ b/index.js
@@ -47,6 +47,13 @@ function Router (opts) {
     this.onBadUrl = null
   }
 
+  if (opts.buildPrettyMeta) {
+    assert(typeof opts.buildPrettyMeta === 'function', 'buildPrettyMeta must be a function')
+    this.buildPrettyMeta = opts.buildPrettyMeta
+  } else {
+    this.buildPrettyMeta = defaultBuildPrettyMeta
+  }
+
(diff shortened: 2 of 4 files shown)


mcollina (changes requested):
Can you make this generic so other metadata could potentially be passed in? fnw know nothing about hooks.

polymathca:
For sure. Would it be better to switch it to a single `includeMeta` option that takes an array of items to pull from the `handler.store` object? We could include a helper on the `fastify` side to make it less cumbersome (e.g. `fastify.printRoutes({ includeHooks: true })` could pass the `Hooks.supportedHooks` object to fmw)

mcollina:
> For sure. Would it be better to switch it to a single `includeMeta` option that takes an array of items to pull from the `route.store` object? 

I think so, yes. It could also be a function to extract such data.

> We could include a helper on the `fastify` side to make it less cumbersome (e.g. `fastify.printRoutes({ includeHooks: true })` could pass the `Hooks.supportedHooks` object to fmw)

Yes exactly.

lib/pretty-print.js
```diff
@@ -10,7 +10,37 @@ const pathDelimiter       = '/'
...
+function buildMetaObject (route, metaArray) {
+  const out = {}
+  if (!Array.isArray(metaArray)) metaArray = route.store ? Object.keys(route.store) : []
+  metaArray.forEach(m => {
+    if (route.store && route.store[m]) {
+      const curr = route.store[m]
+      if (Array.isArray(curr)) {
+        out[m] = curr.map(meta => {
+          if (typeof meta === 'function') return parseFunctionName(meta)
+          return meta
```
Eomm, line 30: This function could be called at line 32 to avoid repetition

test/pretty-print.test.js
```diff
@@ -238,3 +238,176 @@ test('pretty print commonPrefix - handle constrained routes', t => {
...
+
+  const findMyWay = FindMyWay()
+  const namedFunction = () => {}
+  const store = {
+    onRequest: [() => {}, namedFunction],
+    onTimeout: [() => {}],
+    genericMeta: 'meta',
+    mixedMeta: ['mixed items', { an: 'object' }],
+    objectMeta: { one: '1', two: 2 },
+    functionMeta: namedFunction
```
Eomm, line 253: Could you add a test with a `Symbol` too?

Eomm (commented):
Thanks for this PR!

lib/pretty-print.js
polymathca: Good catch, that was getting a little muddy, I moved the type-checking into its own recursive function for sanity

test/pretty-print.test.js
polymathca: Yup! I added support in the bundling function for symbols as both keys and values (and the appropriate tests)

mcollina (approved):
lgtm

polymathca:
I was working on updated tests for the `fastify` integration and I found a bug where bound functions (e.g. hooks from `fastify` routes) weren't returning the correct value. Fixed and tested!

Eomm (approved):
Great!

mcollina (approved):
lgtm

delvedor (changes requested):
I really like the proposal, but I fear it could backfire. This works because it relies on the internal data structure Fastify is using for tracking the state. That structure is not public and might change anytime, causing weird issues here.

polymathca:
> 
> 
> I really like the proposal, but I fear it could backfire. This works because it relies on the internal data structure Fastify is using for tracking the state. That structure is not public and might change anytime, causing weird issues here.

This is getting populated from the `store` object (i.e. `findMyWay.on('get', '/route', (req, res, params, store) => {}, { key: 'value' }`) passed to a route within `find-my-way`. It is then displaying key-value pairs with a bit of parsing to ensure they're logically represented as strings in the `JSON.stringify()` call (e.g. a `Symbol()` explicitly has `.toString()` called, and `function` names are cleaned up)

I have a PR on `fastify` with a convenience shortcut option in the `fastify.printRoutes({ includeHooks: true })` invocation to pass forward an array of the hooks that `fastify` supports, but I think that's the only place where anything `fastify`-specific remains.

Is that `store` functionality likely to change (on either the `fastify` or `find-my-way` side) in a way that might break this approach?

delvedor:
The store is the Fastify internal `Context` object. Currently, that object is a mix of public and private properties, and it will likely change in v4. In any case, Fastify will store private objects in find-my-way, which could change at any time.

See https://github.com/fastify/fastify/pull/2989 and https://github.com/fastify/fastify/blob/6cdd9e1933605da6259484fecdec48166741aa23/lib/route.js#L252.

Probably we can figure out a better strategy in v4, and create a custom internal object for improving the debugging experience, along with the effort of refactoring the errors.

mcollina:
I think the best approach is to make `buildMeta` a function that is specified from the outside during FMW creation. I think this would remove any dependency from FMW to Fastify.

polymathca:
@delvedor Ah, I see what you're saying! Thanks for the links.

So, in reading that, I think that most of the display and object-parsing logic is still good in here (although some variables should be renamed for clarity), but it's more of a matter of deciding how to ensure we're getting a well-formed object to display? I like @mcollina's suggestion of specifying a function on the FMW invocation, which could return an object that this could consume. E.g.

```js
  const router = new FindMyWay({
    buildMeta: (handler) => {
      // using current fastify route structure as example
      const cleanMeta = Object.assign({}, handler.store)

      /* clean up private data */
      delete cleanMeta['privateProperty']

      return cleanMeta
    }
  })
```

with the default just returning the `handler.store` object. Then that could be attached to each handler as a `getMeta()` call, which would only build this object when requested to keep overhead low?

Maybe `Symbol()` keys should be stripped by default, unless they're specifically requested, if that's how `fastify` is keeping items private (although, if this is strictly expose as a getter, maybe having the entire list visible is useful for debugging).

If this is outside the scope of this PR, I can revise the documentation and rename the variables so this feature is just an option to display the `store` object (which I do still feel is useful in the short term) and open a new one to address the `buildMeta` addition.

Eomm:
I think your @polymathca approach is right:

- a buildMeta function on the constructor (maybe `buildPrettyMeta` would be more clear?)
- when the user call prettyPrint the `buildPrettyMeta` (if any) will process the handler, so it will be up to the Fastify source provide this function 

I would really love this feature in prod in v3

polymathca:
> 
> 
> I think your @polymathca approach is right:
> 
>* a buildMeta function on the constructor (maybe `buildPrettyMeta` would be more clear?)
> 
>* when the user call prettyPrint the `buildPrettyMeta` (if any) will process the handler, (what output should return this function?) so it will be up to the Fastify source provide this function
> 
> 
> I would really love this feature in prod in v3

I'd suggest having the `buildPrettyMeta` function output a plain object. That would allow it to work as-is with the current version of `fastify` and still respect the current method of using `route.store` objects in `fmw` for compatibility with any exiting projects using it for routing. Internally, the default could look pretty much like the code I posted earlier:

```js
  FindMyWay.buildPrettyMeta = function(route) {    
    const routeMeta = route.store // this is based on the current behaviour of fmw
    const cleanMeta = Object.assign({}, routeMeta)

    return cleanMeta
  }
```

This function would be overridden if passed a `buildPrettyMeta` function on init. Internally, the `prettyPrint` function would call `buildPrettyMeta` with each `route` object while building the tree output.

`fastify` may need a default `buildPrettyMeta` function to pass when creating its internal router which cleans out any private objects, although I would advocate for allowing the entire `route.store` / `context` object to pass through for ease of debugging.

Does that seem future-proof enough?

mcollina:
that would be perfect!

mcollina:
Is this good to review?

README.md
```diff
@@ -96,6 +96,37 @@ const router = require('find-my-way')({
 })
 ```
 
+You can assign a `buildPrettyPrint` function to sanitize a route's `store` object to use with the `prettyPrint` functions. This function should accept a single object and return an object.
+
+```js
+
+const privateKey = new Symbol('private key')
+const store = { token: '12345' }
```
Eomm, line 104: writing on the declaration seems more clear to me
```suggestion
const store = { token: '12345', [privateKey]: 'private value' }
```

README.md
```diff
@@ -96,6 +96,37 @@ const router = require('find-my-way')({
...
+const router = require('find-my-way')({
+  buildPrettyMeta: route => {
+    const cleanMeta = Object.assign({}, route.store)
+
+    // remove private properties
+    Object.keys(cleanMeta).forEach(k => {
+      if (typeof k === 'symbol') delete cleanMeta[k]
+    })
+
+    return cleanMeta
```
Eomm, line 115:
```suggestion
    return cleanMeta // this will show up in the pretty print output!
```

README.md
```diff
@@ -409,8 +440,9 @@ router.find('GET', '/example', { host: 'fastify.io', version: '1.x' })
 ```
 
 <a name="pretty-print"></a>
-#### prettyPrint([{ commonPrefix: false }])
+#### prettyPrint([{ commonPrefix: false, includeMeta: true || [] }])
```
Eomm, line 443: The `includeMeta` should be removed if I have understood

index.js
```diff
@@ -47,6 +47,13 @@ function Router (opts) {
     this.onBadUrl = null
   }
 
+  if (opts.buildPrettyMeta) {
+    assert(typeof opts.buildPrettyMeta === 'function', 'buildPrettyMeta must be a function')
+    this.buildPrettyMeta = opts.buildPrettyMeta
+  } else {
+    this.buildPrettyMeta = defaultBuildPrettyMeta
```
Eomm, line 54: As default, can we skip it instead of defining an empty default?

Eomm (commented):
Added some thought to clean the code

I'm really excited about this PR ❤️

README.md
polymathca: The `includeMeta` option is what triggers the display, otherwise just the plain tree of routes is returned. It also allows for an array of keys to include (e.g. if you're just interested in seeing `onRequest` hooks, you can pass `{includeMeta:['onRequest']}` to only show that property if it is set on a route.

index.js
polymathca: I added a `defaultBuildPrettyMeta` function at the end of `index.js` which does a basic check to ensure a `route` has been passed, then checks for a store object. It lets a user get up and running with it right away without needing to create their own function to parse and return the store.

mcollina:
still LGTM from me... @delvedor?

polymathca:
> 
> 
> Is this good to review?

Yes, I think it's ready for a look. I added tests and applied a few documentation suggestions @Eomm made.

index.js
```diff
@@ -588,7 +595,7 @@ Router.prototype.prettyPrint = function (opts = {}) {
 
   compressFlattenedNode(root)
 
-  return prettyPrintFlattenedNode(root, '', true)
+  return prettyPrintFlattenedNode(root, '', true, opts, this)
```
Eomm, line 598:
```suggestion
  return prettyPrintFlattenedNode.call(this, root, '', true, opts)
```

lib/pretty-print.js
```diff
@@ -146,7 +183,9 @@ function drawBranch (pathSeg, prefix, endBranch, noPrefix, rootBranch) {
   return branch
 }
 
-function prettyPrintFlattenedNode (flattenedNode, prefix, tail) {
+function prettyPrintFlattenedNode (flattenedNode, prefix, tail, opts, _fmw) {
```
Eomm, line 186: The `_fmw` is required and we should move the function into the Router prototype or use `function.call(context)`

```suggestion
function prettyPrintFlattenedNode (flattenedNode, prefix, tail, opts) {
```

README.md
Eomm: I have understood. Is not it a shortcut to skip the `buildPrettyMeta` function?

index.js
Eomm: thks for clarification 👍🏽

index.js
```diff
@@ -650,3 +657,10 @@ function getClosingParenthensePosition (path, idx) {
 
   throw new TypeError('Invalid regexp expression in "' + path + '"')
 }
+
+function defaultBuildPrettyMeta (route) {
+  // buildPrettyMeta function must return an object, which will be parsed into key/value pairs for display
+  if (!route) return {}
+  if (!route.store) return {}
+  return Object.assign({}, route.store)
```
Eomm, line 665: Is this shallow copy needed?

polymathca: It seems like a good idea since, specifically in the `fastify` use case, we're exposing some properties that are private and could break things if they're mutated. A deep copy felt like overkill since the function can be overridden by the user.

polymathca: > Is not it a shortcut to skip the `buildPrettyMeta` function?

No, the route is forced to go through the `buildPrettyMeta` function before it's displayed. If a user needed to bypass the function, they could override it and just return the original `route` object with no processing.

Eomm (approved):
Great!

README.md
```diff
@@ -96,6 +96,37 @@ const router = require('find-my-way')({
 })
 ```
 
+You can assign a `buildPrettyPrint` function to sanitize a route's `store` object to use with the `prettyPrint` functions. This function should accept a single object and return an object.
```
delvedor, line 99: `buildPrettyPrint` or `buildPrettyMeta`?

delvedor (commented):
LGTM, just a nit

README.md
polymathca: Whoops, good catch! Should be `buildPrettyMeta` (fixed now)

delvedor (approved):

PR #188, opened 2021-04-19
add content negotiation constraint

matthyk:
This is still WIP.

The current plan is to write individual contraints for the different HTTP methods. The advantage of this is that the logic of the individual contraints is significantly reduced and should also be more performant.
Currently, `negotiator` is an additional dependency. This could also be removed by documenting that you have to install it yourself when using these constraints. 
I have also given the user the option to register a `406 handler` that will be returned if none of the requested media types are acceptable.

Are you open to this?

### DIFF
--- a/lib/constrainer.js
+++ b/lib/constrainer.js
@@ -2,13 +2,15 @@
 
 const acceptVersionStrategy = require('./strategies/accept-version')
 const acceptHostStrategy = require('./strategies/accept-host')
+const contentNegotiationForGet = require('./strategies/content-negotiation/content-negotiation.get')
 const assert = require('assert')
 
 class Constrainer {
   constructor (customStrategies) {
     this.strategies = {
       version: acceptVersionStrategy,
-      host: acceptHostStrategy
+      host: acceptHostStrategy,
+      negotiationGet: contentNegotiationForGet
     }
 
     this.strategiesInUse = new Set()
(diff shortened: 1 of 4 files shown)


mcollina:
I would prefer we make this pluggable instead of keep adding strategies here. I do not want to add too many dependency we might not needz

matthyk:
By pluggable do you mean that this would be more of a use case for a Fastify plugin?

mcollina:
I mean you should be able to create a custom constrain strategy: https://github.com/delvedor/find-my-way#custom-constraint-strategies rather than adding it to find-my-way. I don't understand why you need to add this to fmw.

matthyk:
I find it developer-friendly when features are directly integrated and do not have to be installed additionally. But I agree with you that it doesn't fit the minimal approach of fmw. I will focus on just developing my own strategy. Thanks for the feedback.

eladchen:
@matthyk I'm interested in learning what strategy did you choose?

I think it quite essential with a REST API.

Anyway for use to chat so we can share ideas?

matthyk:
@eladchen Since fastify/fastify#3908 it is possible to add an FMW constraint strategy outside the constructor. As a result, it is now possible to develop a Fastify plugin that also adds the necessary route handlers needed if you want to implement content negotation RFC compliant.
Feel free to message me on the Fastify Discord.

PR #193, opened 2021-06-11
Implement '.' separator in parametric routes

### LINKED ISSUE #192: Route with a . (period) between parameters
I've looked over the tests here: https://github.com/delvedor/find-my-way/blob/master/test/issue-17.test.js and I they all seem to check that parameters are successfully parsed when separated for a hyphen. 

Is there a solution for a route like this? `/features/:table/:id.:format` I get the following for the parameters:

```js
{
    "table": "table",
    "id.:format": "1.geojson"
}
```

MSE99:
Closes #192 

Implements `.` as a separator between route params (similar to how `-` works), PR also refactors all `-` tests to test for `.` too.

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,6 +5,7 @@
     '#': 35
     '*': 42
     '-': 45
+    '.': 46
     '/': 47
     ':': 58
     ';': 59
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

delvedor (approved):
LGTM

@mcollina not sure why you marked this as bug, it looks line an enhancement to me.

PR #195, opened 2021-06-26
Migrate to using tap version 15

MSE99:
#### Description

This PR fully migrates find-my-way to use tap version 15, the latest version has a smaller API and supports node >= 10.


mcollina (approved):
lgtm

PR #197, opened 2021-07-06
Fix mustMatchWhenDerived for route nodes with both unconstrained and constrained routes 

### LINKED ISSUE #196: mustMatchWhenDerived ignored
The `mustMatchWhenDerived` flag seems ignored configuring a custom constraint.

Starting from this example, the last lookup with `hello: C` header, it should not execute the `NO CONSTRAIN` handler for what I have understood by this flag:

> if handlers without constraints can match requests that have a value for this constraint

I understand: if `mustMatchWhenDerived` is true, a handler without constraint is not considered.

```js
const FindMyWay = require('find-my-way')

const router = FindMyWay({
  constraints: {
    hello: {
      name: 'hello',
      mustMatchWhenDerived: true,
      storage: function strictStringStorage () {
        const holder = new Map()
        return {
          get (value) {
            return holder.get(value)
          },
          set (value, store) {
            holder.set(value, store)
          },
          del (value) {
            holder.delete(value)
          },
          empty () {
            holder.clear()
          }
        }
      },
      validate (value) {
        //
      },
      deriveConstraint: (req, ctx) => { return req.headers.hello }
    }
  }
})

router.on('GET', '/', (req, res) => {
  console.log('NO CONSTRAINT')
})

router.on('GET', '/', { constraints: { hello: 'A' } }, (req, res) => {
  console.log('A')
})

router.on('GET', '/', { constraints: { hello: 'B' } }, (req, res) => {
  console.log('B')
})

router.lookup({ method: 'GET', url: '/', headers: { } }, null)
router.lookup({ method: 'GET', url: '/', headers: { hello: 'A' } }, null)
router.lookup({ method: 'GET', url: '/', headers: { hello: 'B' } }, null)
router.lookup({ method: 'GET', url: '/', headers: { hello: 'C' } }, null)  // print NO CONSTRAINT but I would expect `null`

```

NB: the output is always the same, regardless of the flag 😞 

cc @airhorns

airhorns:
Fixes #196

mustMatchWhenDerived strategies ensure that requests that pass a value for the constraint (like a version number) should only match routes that are constrained using that strategy. This avoids invoking unconstrained, unversioned routes for requests that are versioned. find-my-way has supported this for a while, but all the work to add custom constraint strategies broke this support in a case where both unconstrained and constrained routes were added for the same path. This fixes the tests to exercise that case and ensures the fast constraint matcher system doesn't allow unconstrained routes to match the mustMatchWhenDerived strategies!

This also fixes a bug where mustMatchWhenDerived constraints need to be used by a route before they start taking effect. 
For these, we need to ensure we always derive a value for the constraint even if no routes are using that constraint yet. That way, we know if the route has a value that must match. This was a bug caused by the performance optimization of trying to avoid deriving constraint values for constraints that aren't in use -- we must consider all custom mustMatchWhenDerived constraints in use in order to check if the value is present.

### DIFF
--- a/node.js
+++ b/node.js
@@ -310,7 +310,11 @@ Node.prototype._compileGetHandlerMatchingConstraints = function () {
   }
   // Return the first handler who's bit is set in the candidates https://stackoverflow.com/questions/18134985/how-to-find-index-of-first-set-bit
   lines.push(`
-  return this.handlers[Math.floor(Math.log2(candidates))]
+  const handler = this.handlers[Math.floor(Math.log2(candidates))]
+  if (handler && derivedConstraints.__hasMustMatchValues && handler === this.unconstrainedHandler) {
+    return null;
+  }
+  return handler;
   `)
 
   this._getHandlerMatchingConstraints = new Function('derivedConstraints', lines.join('\n')) // eslint-disable-line
--- a/lib/constrainer.js
+++ b/lib/constrainer.js
@@ -24,6 +24,10 @@ class Constrainer {
         assert(strategy.deriveConstraint && typeof strategy.deriveConstraint === 'function', 'strategy.deriveConstraint function is required.')
         strategy.isCustom = true
         this.strategies[strategy.name] = strategy
+
+        if (strategy.mustMatchWhenDerived) {
+          this.noteUsage({ [kCustomStrategies[i]]: strategy })
+        }
       }
     }
   }
--- a/test/constraint.custom.test.js
+++ b/test/constraint.custom.test.js
@@ -20,7 +20,7 @@ const customHeaderConstraint = {
     }
   },
   deriveConstraint: (req, ctx) => {
-    return req.headers.accept
+    return req.headers['user-agent']
   }
 }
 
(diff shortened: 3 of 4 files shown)


node.js
```diff
@@ -310,7 +310,11 @@ Node.prototype._compileGetHandlerMatchingConstraints = function () {
   }
   // Return the first handler who's bit is set in the candidates https://stackoverflow.com/questions/18134985/how-to-find-index-of-first-set-bit
   lines.push(`
-  return this.handlers[Math.floor(Math.log2(candidates))]
+  const handler = this.handlers[Math.floor(Math.log2(candidates))]
+  if (handler && derivedConstraints.__hasMustMatchValues && handler == this.unconstrainedHandler) {
```
Eomm, line 314:
```suggestion
  if (handler && derivedConstraints.__hasMustMatchValues && handler === this.unconstrainedHandler) {
```

Eomm (commented):
Thanks a lot for this PR 🚀

mcollina (approved):
lgtm

Eomm (approved):
🔝

delvedor (approved):
LGTM

PR #200, opened 2021-07-09
Add Typings for `prettyPrint` with Options

lusbuab:


mcollina (changes requested):
Can you please add a unit test? we use `tsd`.

mcollina (approved):
lgtm

PR #207, opened 2021-10-06
fix decoding url

### LINKED ISSUE #206: the lib doesn't handle url-encoded URI properly
### The problem

The lib doesn't handle url-encoded URI properly. It's impossible to set tricky routes. There are many issues are related to this root reason ([fastify-static-143](https://github.com/fastify/fastify-static/issues/143), [fastify-static-234](https://github.com/fastify/fastify-static/issues/234), [fastify-3356](https://github.com/fastify/fastify/issues/3356)).

### The example

Let's assume that we want to serve 2 files:
1. `./[...]/a .md`
2. `./[...]/a%20.md`

To do so we create two routes:
```js
const http = require('http')

router.on('GET', '/[...]/a .md', (req, res, params) => {
    res.end('serve 1st file with space')
})
router.on('GET', '/[...]/a%20.md', (req, res, params) => {
    res.end('serve 2nd file with %')
})

    router.lookup(req, res)
})

server.listen(3000)
```

The 1st route is not reachable at all.
The 2nd route is reachable only if we try to get to the 1st route.

This is very confusing.

There is more to this:
The browser (at least now) doesn't encode `[` and `]`. But it should be encoded to support IPv6.
Let's assume we create a client app which will consume the files. We will definitely use `encodeURI` before we send the request.
In this case we'll send `/%5B...%5D/a%20.md` as the URI and we will never reach the route.

I want to highlight here that `encodeURI` before setup a route in the example **will not solve the problem** as well, it just makes things worse instead.

### The theory

The URI is restricted to ASCII only. To **transport** the unicode character over ASCII we must encode it in percent-notation (`µ` -> `%C2%B5`).
Besides unicode characters there are special characters like: ?, #, ;, ... It is also possible to have those in URI but they also must be encoded  in percent-notation (` ` -> `%20`)

According to [rfc2616](https://datatracker.ietf.org/doc/html/rfc2616#section-3.2.3):
```
      the following three URIs are equivalent:

      http://abc.com:80/~smith/home.html
      http://ABC.com/%7Esmith/home.html
      http://ABC.com:/%7esmith/home.html
```

Based on listed above we must convert all the variants of the URI to a single final format, in other words we must `decodeURI` it.
It must be done right after we get the request because the encoding was done only for **transport** purposes.

### The solution

Use `decodeURI` somewhere [here](https://github.com/delvedor/find-my-way/blob/c561d220e6c0c93b76b72a7461dcdc8620158ba2/index.js#L344) and never use `encodeURI` while setting up the routes.

Eomm:
Fixes #206

### DIFF
--- a/README.md
+++ b/README.md
@@ -315,7 +315,7 @@ In this case as parameter separator it's possible to use whatever character is n
 
 Having a route with multiple parameters may affect negatively the performance, so prefer single parameter approach whenever possible, especially on routes which are on the hot path of your application.
 
-**Note** that the you must encode those parameters that contains [reserved characters](https://www.rfc-editor.org/rfc/rfc3986#section-2.2).
+**Note** that the you must encode parameters that contain [reserved characters](https://www.rfc-editor.org/rfc/rfc3986#section-2.2).
 
 <a name="match-order"></a>
 ##### Match order
--- a/README.md
+++ b/README.md
@@ -315,6 +315,8 @@ In this case as parameter separator it's possible to use whatever character is n
 
 Having a route with multiple parameters may affect negatively the performance, so prefer single parameter approach whenever possible, especially on routes which are on the hot path of your application.
 
+**Note** that the you must encode those parameters that contains [reserved characters](https://www.rfc-editor.org/rfc/rfc3986#section-2.2).
+
 <a name="match-order"></a>
 ##### Match order
 
--- a/README.md
+++ b/README.md
@@ -315,7 +315,7 @@ In this case as parameter separator it's possible to use whatever character is n
 
 Having a route with multiple parameters may affect negatively the performance, so prefer single parameter approach whenever possible, especially on routes which are on the hot path of your application.
 
-**Note** that the you must encode parameters that contain [reserved characters](https://www.rfc-editor.org/rfc/rfc3986#section-2.2).
+**Note** that you must encode the parameters containing [reserved characters](https://www.rfc-editor.org/rfc/rfc3986#section-2.2).
 
 <a name="match-order"></a>
 ##### Match order
--- a/bench.js
+++ b/bench.js
@@ -79,7 +79,10 @@ suite
   .add('find dynamic route', function () {
     findMyWay.find('GET', '/user/tomas', undefined)
   })
-  .add('find dynamic route with encoded parameter', function () {
+  .add('find dynamic route with encoded parameter unoptimized', function () {
+    findMyWay.find('GET', '/user/maintainer%2Btomas', undefined)
+  })
+  .add('find dynamic route with encoded parameter optimized', function () {
     findMyWay.find('GET', '/user/maintainer%20tomas', undefined)
   })
   .add('find dynamic multi-parametric route', function () {
@@ -97,8 +100,11 @@ suite
   .add('find long nested dynamic route', function () {
     findMyWay.find('GET', '/posts/10/comments/42/author', undefined)
   })
-  .add('find long nested dynamic route with encoded parameter', function () {
+  .add('find long nested dynamic route with encoded parameter unoptimized', function () {
     findMyWay.find('GET', '/posts/10%2C10/comments/42%2C42/author', undefined)
+  })
+  .add('find long nested dynamic route with encoded parameter optimized', function () {
+    findMyWay.find('GET', '/posts/10%2510/comments/42%2542/author', undefined)
   })
   .add('find long nested dynamic route with other method', function () {
     findMyWay.find('POST', '/posts/10/comments', undefined)
--- a/bench.js
+++ b/bench.js
@@ -79,6 +79,9 @@ suite
   .add('find dynamic route', function () {
     findMyWay.find('GET', '/user/tomas', undefined)
   })
+  .add('find dynamic route with encoded parameter', function () {
+    findMyWay.find('GET', '/user/maintainer%20tomas', undefined)
+  })
   .add('find dynamic multi-parametric route', function () {
     findMyWay.find('GET', '/customer/john-doe', undefined)
   })
@@ -93,6 +96,9 @@ suite
   })
   .add('find long nested dynamic route', function () {
     findMyWay.find('GET', '/posts/10/comments/42/author', undefined)
+  })
+  .add('find long nested dynamic route with encoded parameter', function () {
+    findMyWay.find('GET', '/posts/10%2C10/comments/42%2C42/author', undefined)
   })
   .add('find long nested dynamic route with other method', function () {
     findMyWay.find('POST', '/posts/10/comments', undefined)
--- a/index.js
+++ b/index.js
@@ -2,15 +2,27 @@
 
 /*
   Char codes:
+    '!': 33 - !
     '#': 35 - %23
+    '$': 36 - %24
     '%': 37 - %25
-    '*': 42
-    '-': 45
-    '.': 46
+    '&': 38 - %26
+    ''': 39 - '
+    '(': 40 - (
+    ')': 41 - )
+    '*': 42 - *
+    '+': 43 - %2B
+    ',': 44 - %2C
+    '-': 45 - -
+    '.': 46 - .
     '/': 47 - %2F
     ':': 58 - %3A
-    ';': 59
+    ';': 59 - %3B
+    '=': 61 - %3D
     '?': 63 - %3F
+    '@': 64 - %40
+    '_': 95 - _
+    '~': 126 - ~
 */
 
 const assert = require('assert')
@@ -366,7 +378,6 @@ Router.prototype.find = function find (method, path, derivedConstraints) {
       ? this._onBadUrl(path)
       : null
   }
-  // console.log(sanitizedPath)
 
   var originalPath = path
   var originalPathLength = path.length
@@ -493,7 +504,10 @@ Router.prototype.find = function find (method, path, derivedConstraints) {
 
     // wildcard route
     if (kind === NODE_TYPES.MATCH_ALL) {
-      decoded = fastDecode(originalPath.slice(idxInOriginalPath))
+      decoded = sanitizedPath.containsEncodedComponents
+        ? fastDecode(sanitizedPath.originPath.slice(idxInOriginalPath))
+        : originalPath.slice(idxInOriginalPath)
+
       if (decoded === null) {
         return this.onBadUrl !== null
           ? this._onBadUrl(originalPath.slice(idxInOriginalPath))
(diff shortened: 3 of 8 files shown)


mav-rik:
@Eomm I would rather put decodeURI right at sanitizeUrl output.

First of all decodeURI must be wrapped to be called safe: 
```js
function safeDecodeURI (v) {
    if (v.indexOf('%') < 0) return v
    try {
        return decodeURI(v)
    } catch (e) {
        return null
    }
}
```
and the `sanitizeUrl`:
```js
function sanitizeUrl (url) {
  for (var i = 0, len = url.length; i < len; i++) {
    var charCode = url.charCodeAt(i)
    // Thus, we need to split on `;` as well as `?` and `#`.
    if (charCode === 63 || charCode === 59 || charCode === 35) {
      return safeDecodeURI(url.slice(0, i))
    }
  }
  return safeDecodeURI(url)
}
```

All the calls of fastDecode should stay untached for further decoding of dynamic params.

Eomm:
> we should sanitize only once and the find method is recursive.

This was not true: the iteration happens inside the `while` loop:

https://github.com/delvedor/find-my-way/blob/c561d220e6c0c93b76b72a7461dcdc8620158ba2/index.js#L375

so the `sanitizeUrl` is executed once

Eomm:
> I'm not liking the decreased performance, but that's expectable.

Yes. I want to filter out cases like the space char `%20` because in this case, we can skip `decodeURIComponent`
We should run only the code strictly needed for reducing the performance drop

Eomm:
New code update and new benches

<details>
  <summary>master</summary>

```diff
lookup static route x 78,133,409 ops/sec ±0.29% (594 runs sampled)
lookup dynamic route x 6,649,181 ops/sec ±0.09% (597 runs sampled)
lookup dynamic multi-parametric route x 3,526,789 ops/sec ±0.12% (595 runs sampled)
lookup dynamic multi-parametric route with regex x 2,709,928 ops/sec ±0.12% (595 runs sampled)
lookup long static route x 5,196,753 ops/sec ±0.11% (594 runs sampled)
lookup long dynamic route x 3,931,994 ops/sec ±0.11% (596 runs sampled)
lookup static route on constrained router x 25,913,319 ops/sec ±0.44% (584 runs sampled)
lookup static versioned route x 4,246,000 ops/sec ±0.24% (594 runs sampled)
lookup static constrained (version & host) route x 4,247,483 ops/sec ±0.13% (596 runs sampled)
find static route x 53,589,163 ops/sec ±0.41% (591 runs sampled)
find dynamic route x 7,220,519 ops/sec ±0.29% (595 runs sampled)
+find dynamic route with encoded parameter unoptimized x 4,769,694 ops/sec ±0.42% (595 runs sampled)
+find dynamic route with encoded parameter optimized x 4,913,347 ops/sec ±0.14% (596 runs sampled)
find dynamic multi-parametric route x 3,893,650 ops/sec ±0.09% (596 runs sampled)
find dynamic multi-parametric route with regex x 2,934,077 ops/sec ±0.11% (597 runs sampled)
find long static route x 7,297,113 ops/sec ±0.11% (596 runs sampled)
find long dynamic route x 5,261,343 ops/sec ±0.17% (596 runs sampled)
find long nested dynamic route x 2,149,705 ops/sec ±0.11% (596 runs sampled)
+find long nested dynamic route with encoded parameter unoptimized x 1,718,396 ops/sec ±0.08% (596 runs sampled)
+find long nested dynamic route with encoded parameter optimized x 1,748,907 ops/sec ±0.09% (596 runs sampled)
find long nested dynamic route with other method x 4,416,712 ops/sec ±0.10% (596 runs sampled)
find long nested dynamic route x 2,122,985 ops/sec ±0.10% (595 runs sampled)
find long nested dynamic route with other method x 4,423,109 ops/sec ±0.09% (595 runs sampled)
```
</details>


<details>
  <summary>this PR</summary>

```diff
lookup static route x 76,436,594 ops/sec ±0.12% (595 runs sampled)
lookup dynamic route x 6,665,876 ops/sec ±0.16% (595 runs sampled)
lookup dynamic multi-parametric route x 3,498,919 ops/sec ±0.12% (595 runs sampled)
lookup dynamic multi-parametric route with regex x 2,712,085 ops/sec ±0.10% (595 runs sampled)
lookup long static route x 4,949,353 ops/sec ±0.10% (596 runs sampled)
lookup long dynamic route x 3,900,072 ops/sec ±0.10% (595 runs sampled)
lookup static route on constrained router x 24,607,394 ops/sec ±0.21% (588 runs sampled)
lookup static versioned route x 4,185,986 ops/sec ±0.10% (596 runs sampled)
lookup static constrained (version & host) route x 4,197,256 ops/sec ±0.10% (596 runs sampled)
find static route x 50,484,083 ops/sec ±0.56% (583 runs sampled)
find dynamic route x 6,635,193 ops/sec ±0.33% (593 runs sampled)
-find dynamic route with encoded parameter unoptimized x 1,753,345 ops/sec ±0.22% (593 runs sampled)
-find dynamic route with encoded parameter optimized x 1,918,291 ops/sec ±0.17% (593 runs sampled)
find dynamic multi-parametric route x 3,250,562 ops/sec ±0.25% (594 runs sampled)
find dynamic multi-parametric route with regex x 2,622,539 ops/sec ±0.38% (597 runs sampled)
find long static route x 4,599,185 ops/sec ±0.08% (597 runs sampled)
find long dynamic route x 3,546,332 ops/sec ±0.11% (597 runs sampled)
find long nested dynamic route x 1,901,969 ops/sec ±0.17% (596 runs sampled)
-find long nested dynamic route with encoded parameter unoptimized x 916,412 ops/sec ±0.10% (595 runs sampled)
-find long nested dynamic route with encoded parameter optimized x 964,830 ops/sec ±0.12% (594 runs sampled)
find long nested dynamic route with other method x 3,758,202 ops/sec ±0.10% (596 runs sampled)
find long nested dynamic route x 1,869,098 ops/sec ±0.18% (595 runs sampled)
find long nested dynamic route with other method x 3,744,556 ops/sec ±0.13% (596 runs sampled)
```
</details>

mcollina:
How are we with benchmarks?

Eomm:
here the results
the encoded parameter benches have a -50% performance drop

<details>
  <summary>comparing master-pr</summary>

```diff
  pr:     lookup static route x 76,674,656 ops/sec ±0.12% (595 runs sampled)
+ master: lookup static route x 79,131,796 ops/sec ±0.25% (594 runs sampled)

  pr:     lookup dynamic route x 6,259,776 ops/sec ±0.29% (596 runs sampled)
+ master: lookup dynamic route x 6,604,871 ops/sec ±0.19% (595 runs sampled)

  pr:     lookup dynamic multi-parametric route x 3,508,864 ops/sec ±0.29% (594 runs sampled)
+ master: lookup dynamic multi-parametric route x 3,555,251 ops/sec ±0.11% (595 runs sampled)

  pr:     lookup dynamic multi-parametric route with regex x 2,816,789 ops/sec ±0.46% (596 runs sampled)
+ master: lookup dynamic multi-parametric route with regex x 2,704,261 ops/sec ±0.15% (595 runs sampled)

  pr:     lookup long static route x 4,970,945 ops/sec ±0.11% (595 runs sampled)
+ master: lookup long static route x 5,166,473 ops/sec ±0.19% (593 runs sampled)

  pr:     lookup long dynamic route x 3,939,538 ops/sec ±0.10% (596 runs sampled)
+ master: lookup long dynamic route x 3,884,604 ops/sec ±0.96% (593 runs sampled)

  pr:     lookup static route on constrained router x 24,927,696 ops/sec ±0.20% (588 runs sampled)
+ master: lookup static route on constrained router x 24,286,435 ops/sec ±0.22% (590 runs sampled)

  pr:     lookup static versioned route x 4,206,344 ops/sec ±0.10% (596 runs sampled)
+ master: lookup static versioned route x 4,297,009 ops/sec ±0.17% (595 runs sampled)

  pr:     lookup static constrained (version & host) route x 4,206,567 ops/sec ±0.16% (596 runs sampled)
+ master: lookup static constrained (version & host) route x 4,335,753 ops/sec ±0.20% (596 runs sampled)

  pr:     find static route x 52,060,384 ops/sec ±0.38% (586 runs sampled)
+ master: find static route x 54,845,502 ops/sec ±0.41% (585 runs sampled)

  pr:     find dynamic route x 6,809,845 ops/sec ±0.13% (596 runs sampled)
+ master: find dynamic route x 7,363,324 ops/sec ±0.14% (595 runs sampled)

  pr:     find dynamic route with encoded parameter unoptimized x 1,794,848 ops/sec ±0.13% (596 runs sampled)
+ master: find dynamic route with encoded parameter unoptimized x 4,879,835 ops/sec ±0.15% (596 runs sampled)

  pr:     find dynamic route with encoded parameter optimized x 1,916,581 ops/sec ±0.15% (590 runs sampled)
+ master: find dynamic route with encoded parameter optimized x 4,955,861 ops/sec ±0.12% (596 runs sampled)

  pr:     find dynamic multi-parametric route x 3,347,742 ops/sec ±0.12% (594 runs sampled)
+ master: find dynamic multi-parametric route x 3,957,345 ops/sec ±0.10% (596 runs sampled)

  pr:     find dynamic multi-parametric route with regex x 2,686,346 ops/sec ±0.12% (597 runs sampled)
+ master: find dynamic multi-parametric route with regex x 2,938,081 ops/sec ±0.33% (596 runs sampled)

  pr:     find long static route x 4,479,007 ops/sec ±0.11% (595 runs sampled)
+ master: find long static route x 7,311,289 ops/sec ±0.16% (595 runs sampled)

  pr:     find long dynamic route x 3,392,116 ops/sec ±0.11% (592 runs sampled)
+ master: find long dynamic route x 5,357,651 ops/sec ±0.10% (595 runs sampled)

  pr:     find long nested dynamic route x 1,882,345 ops/sec ±0.12% (595 runs sampled)
+ master: find long nested dynamic route x 2,174,753 ops/sec ±0.09% (596 runs sampled)

  pr:     find long nested dynamic route with encoded parameter unoptimized x 944,112 ops/sec ±0.16% (596 runs sampled)
+ master: find long nested dynamic route with encoded parameter unoptimized x 1,732,654 ops/sec ±0.11% (596 runs sampled)

  pr:     find long nested dynamic route with encoded parameter optimized x 976,013 ops/sec ±0.15% (596 runs sampled)
+ master: find long nested dynamic route with encoded parameter optimized x 1,762,524 ops/sec ±0.11% (596 runs sampled)

  pr:     find long nested dynamic route with other method x 3,700,463 ops/sec ±0.12% (591 runs sampled)
+ master: find long nested dynamic route with other method x 4,476,532 ops/sec ±0.12% (596 runs sampled)

  pr:     find long nested dynamic route x 1,856,434 ops/sec ±0.26% (593 runs sampled)
+ master: find long nested dynamic route x 2,136,509 ops/sec ±0.10% (596 runs sampled)

  pr:     find long nested dynamic route with other method x 3,737,505 ops/sec ±0.12% (596 runs sampled)
+ master: find long nested dynamic route with other method x 4,478,782 ops/sec ±0.10% (595 runs sampled)
```
</details>

darkgl0w:
There is a little typo on the `README.md` file.
It should be :

```markdown
**Note** that you must encode parameters containing [reserved characters](https://www.rfc-editor.org/rfc/rfc3986#section-2.2).
```

README.md
```diff
@@ -315,6 +315,8 @@ In this case as parameter separator it's possible to use whatever character is n
 
 Having a route with multiple parameters may affect negatively the performance, so prefer single parameter approach whenever possible, especially on routes which are on the hot path of your application.
 
+**Note** that the you must encode parameters that contain [reserved characters](https://www.rfc-editor.org/rfc/rfc3986#section-2.2).
```
darkgl0w, line 318:
```suggestion
**Note** that you must encode parameters containing [reserved characters](https://www.rfc-editor.org/rfc/rfc3986#section-2.2).
```

mcollina (approved):
lgtm

mcollina:
@delvedor could you land and ship a new major?

Eomm:
@delvedor I may target a `next` branch if needed 👍🏽

delvedor (approved):

PR #210, opened 2021-11-02
Fix host constraints not working when http2 is enabled

anker2000:
Find-my-way is relying on req.headers.host to determine the host, however when http2 is enabled req.headers.host won't return the host name. For http2 req.headers[':authority'] could be used.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -9,7 +9,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 14.x]
+        node-version: [12.x, 14.x, 16.x, 17.x]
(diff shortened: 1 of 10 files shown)


Eomm:
Could you add unit tests and revert the `package.json` changes?
We will take care of the version

anker2000:
I've updated the package.json file (sorry - forgot to clean that up), but this is my first ever pull request, and I unfortunately have no idea what a unit test is.

mcollina:
https://github.com/delvedor/find-my-way/blob/36e1888a8158cc030e0ab586064d828f5aa8e00f/test/server.test.js < this file test a lot of things for http. You should create one for http2.

mcollina:
Can you rebase? There are spurious commits.

PR #214, opened 2021-11-29
add :authority to host constraint to support http2

luisorbaiceta:
fixes fastify/fastify#3408

### DIFF
--- a/lib/constrainer.js
+++ b/lib/constrainer.js
@@ -92,7 +92,7 @@ class Constrainer {
         if (key === 'version') {
           lines.push('   version: req.headers[\'accept-version\'],')
         } else if (key === 'host') {
-          lines.push('   host: req.headers.host,')
+          lines.push('   host: req.headers.host || req.headers[\':authority\'],')
         } else {
           throw new Error('unknown non-custom strategy for compiling constraint derivation function')
         }
--- a/test/http2/constraint.host.test.js
+++ b/test/http2/constraint.host.test.js
@@ -1,0 +1,45 @@
+'use strict'
+
+const t = require('tap')
+const test = t.test
+const FindMyWay = require('../..')
+
+test('A route supports host constraints under http2 protocol', t => {
+  t.plan(3)
+
+  const findMyWay = FindMyWay()
+
+  findMyWay.on('GET', '/', {}, (req, res) => {
+    t.fail()
+  })
+  findMyWay.on('GET', '/', { constraints: { host: 'fastify.io' } }, (req, res) => {
+    t.equal(req.headers[':authority'], 'fastify.io')
+  })
+  findMyWay.on('GET', '/', { constraints: { host: /.+\.de/ } }, (req, res) => {
+    t.ok(req.headers[':authority'].endsWith('.de'))
+  })
+
+  findMyWay.lookup({
+    method: 'GET',
+    url: '/',
+    headers: {
+      ':authority': 'fastify.io'
+    }
+  })
+
+  findMyWay.lookup({
+    method: 'GET',
+    url: '/',
+    headers: {
+      ':authority': 'fastify.de'
+    }
+  })
+
+  findMyWay.lookup({
+    method: 'GET',
+    url: '/',
+    headers: {
+      ':authority': 'find-my-way.de'
+    }
+  })
+})


mcollina (commented):
If current tests support this, then Fastify should not have a but. Maybe add one more?

luisorbaiceta:
I could add some dev dependecies to create a server and check the router over an http2 connection. But current tests are not doing this, they just verify handlers are equal over registered routes. How do you think we could approach this? @mcollina

mcollina:
Why this would not be enough?

```js
t.equal(findMyWay.find('GET', '/', { ':authority': 'example.com' }).handler, gamma)
```

luisorbaiceta:
`:authority` is not a valid option for the current `Constrainer`

mcollina:
> :authority is not a valid option for the current Constrainer and gets sanitized. To achieve this we would have to create a new constraint strategy and test that the router finds that registered route with the constraint.

I don't understand this. You are modifying a strategy in this PR. This can definitely be tested here.

luisorbaiceta:
I am only modifying the host strategy like this:

`lines.push('   host: req.headers.host || req.headers[\':authority\'],')`

This will only be resolved at request time.

mcollina:
> This will only be resolved at request time.

It will be called at https://github.com/delvedor/find-my-way/blob/70844a9c1effe065915ceabef3867abb1add618d/index.js#L388.

This is where deriveConstraint is built: https://github.com/delvedor/find-my-way/blob/70844a9c1effe065915ceabef3867abb1add618d/lib/constrainer.js#L75-L119.

luisorbaiceta:
Sorry @mcollina, I don't see how this can be tested without creating a server. When trying to test this as suggested:

```js
'use strict'

const t = require('tap')
const test = t.test
const FindMyWay = require('../..')

const alpha = () => { }
const beta = () => { }
const gamma = () => { }

const { buildCertificate } = require('./build-certificate')
t.before(buildCertificate)

test('A route supports host constraints under http2 protocol', t => {
  t.plan(4)

  const findMyWay = FindMyWay()

  findMyWay.on('GET', '/', {}, alpha)
  findMyWay.on('GET', '/', { constraints: { host: 'fastify.io' } }, beta)
  findMyWay.on('GET', '/', { constraints: { host: 'example.com' } }, gamma)

  t.equal(findMyWay.find('GET', '/', {}).handler, alpha)
  t.equal(findMyWay.find('GET', '/', { ':authority': 'something-else.io' }).handler, alpha)
  t.equal(findMyWay.find('GET', '/', { ':authority': 'fastify.io' }).handler, beta)
  t.equal(findMyWay.find('GET', '/', { ':authority': 'example.com' }).handler, gamma)
})
```

Tests fail:

```
diff: |
        --- expected
        +++ actual
        @@ -1,1 +1,1 @@
        -Function beta()
        +Function alpha()
```

This is because the find method is trying to find a matching route with the ':authority' constraint, but this constraint does not exist by default, so either we add it, or we create a custom strategy, which in my opinion does not fit well for our testing purposes

matthyk:
Any changes should also be tested.
You could use the `lookup` method instead.

```js
'use strict'

const t = require('tap')
const test = t.test
const FindMyWay = require('..')

test('A route supports host constraints under http2 protocol', t => {
  t.plan(3)

  const findMyWay = FindMyWay()

  findMyWay.on('GET', '/', {}, (req, res) => {
    t.fail()
  })
  findMyWay.on('GET', '/', { constraints: { host: 'fastify.io' } }, (req, res) => {
    t.equal(req.headers[':authority'], 'fastify.io')
  })
  findMyWay.on('GET', '/', { constraints: { host: /.+\.de/ } }, (req, res) => {
    t.ok(req.headers[':authority'].endsWith('.de'))
  })

  findMyWay.lookup({
    method: 'GET',
    url: '/',
    headers: {
      ':authority': 'fastify.io'
    }
  })

  findMyWay.lookup({
    method: 'GET',
    url: '/',
    headers: {
      ':authority': 'fastify.de'
    }
  })

  findMyWay.lookup({
    method: 'GET',
    url: '/',
    headers: {
      ':authority': 'find-my-way.de'
    }
  })
})
```

luisorbaiceta:
I missed the lookup method, thanks for that @matthyk!

mcollina (approved):
lgtm

mcollina:
Thanks @matthyk!

delvedor (approved):
LGTM

PR #216, opened 2021-12-01
fix ci fail due to node.js name collision on windows

### LINKED ISSUE #215: Lint job in github action not working for Windows environments
CI pipeline is failing due to a  `node.js` name collision in windows systems

luisorbaiceta:
seems that current CI pipeline error is somehow related to a collision with the node.js file.

fixes #215


luisorbaiceta:
> Please restore the previous linting

Done. I didn't notice I had prettier extension in the windows environment sorry

Eomm:
> seems that current CI pipeline error is somehow related to a collision with the node.js file.
> 
> fixes #215

Could you point to some `standard` documentation about it?
I think this issue should be solved on their site too

luisorbaiceta:
Is not a `standard` problem, let me change the title

luisorbaiceta:
I've done some research and It seems that is related to DOS, OS/2, and Windows systems and the way they handle the `PATH` variable.

>When a command is entered in a command shell or a system call is made by a program to execute a program, the system first searches the current working directory and then searches the path, examining each directory from left to right, looking for an executable filename that matches the command name given. Executable programs have filename extensions of EXE or COM, and batch scripts have extensions of BAT or CMD. Other executable filename extensions can be registered with the system as well.

Whereas in Unix and Unix-like the process is different.

>When a command name is specified by the user or an exec call is made from a program, the system searches through $PATH, examining each directory from left to right in the list, looking for a filename that matches the command name. Once found, the program is executed as a child process of the command shell or program that issued the command.

In this case, `poweshell` was trying to execute the `node.js` file in the project scope.

mcollina:
@simoneb could you check out https://github.com/delvedor/find-my-way/pull/216?

simoneb:
I can confirm that without this change, running tests on Windows leads to some very fancy behaviors. With this change, all works good

mcollina (approved):
lgtm

Issue #219, opened 2021-12-20
Double colon route breaks default fallback when there is another route with a common prefix.


emfomy:
I register two routes:

- `GET /hello`
- `GET /hello:world`

Both routes work fine. However, the server freeze when sending request to `/hello:worl`, `/hello:world1`, `/hello:bar`, `/hello123` (i.e. and route start with `/hello`).

Without `GET /hello` registered, these routes can fall back to 404 properly.

### Code

```javascript
const http = require("http");
const router = require("find-my-way")();

router.on("GET", "/hello", (req, res, params) => {
  res.end("hello");
});

router.on("GET", "/hello::world", (req, res, params) => {
  res.end("hello world");
});

const server = http.createServer((req, res) => {
  router.lookup(req, res);
});

server.listen(3000, (err) => {
  if (err) throw err;
  console.log("Server listening on: http://localhost:3000");
});
```

### Related Issue
- fastify/fastify#3550

emfomy:
According to @climba03003 (https://github.com/fastify/fastify/issues/3550#issuecomment-997624300), it is related to the infinite loop in `.find` method.

https://github.com/delvedor/find-my-way/blob/695654cdd452d2993ada90cc87bdf8d695aee076/index.js#L497-L505

emfomy:
Fixed in [v5.1.1](https://github.com/delvedor/find-my-way/releases/tag/v5.1.1).

PR #220, opened 2021-12-20
fix: double colon route issue (#219)

bugagashenkj:
Hi, please don't judge me too hard since it's my first contribution here. I think the problem is that the `parametricBrother` field can be set even if `child.prefix` starts with `":"`. In this case, these are two colons in a row, which should be parsed as a simple path, and not as a path with a parameter. I'm not sure that it's the best solution, but it works, so I decided to create a PR.

### DIFF
--- a/custom_node.js
+++ b/custom_node.js
@@ -63,41 +63,35 @@ Node.prototype.addChild = function (node) {
   )
 
   this.children[label] = node
-  this.numberOfChildren = Object.keys(this.children).length
 
-  const labels = Object.keys(this.children)
-  var parametricBrother = this.parametricBrother
-  for (var i = 0; i < labels.length; i++) {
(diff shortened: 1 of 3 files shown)


RafaelGSS (changes requested):
Would you mind adding tests to your fix?

RafaelGSS (approved):
Great work!

bugagashenkj:
Fixed #222. I added changes to this PR because they were fixed in the same place.

mcollina (approved):
lgtm

mcollina:
we are waiting for a review & release from @delvedor

delvedor (approved):
LGTM

delvedor:
Released in `v5.1.1` 🚀

Issue #222, opened 2021-12-23
Incorrect parametric brother route set up.


bugagashenkj:
Sometimes it sets a non-nearest parametric brother route. Static route `/text/hello` will have `/:c` as a parametric brother, but should have `text/:e/test`.

It's not the same issue with (#221)

```js
router.on('GET', '/text/hello', () => {})
router.on('GET', '/text/:e/test', () => {})
router.on('GET', '/:c', () => {})

assert.equal(router.find('GET', '/text/hellos/test'), null) // shouldn't be null
```

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

bugagashenkj:
If you set up routes in a different order this will work
```js
router.on('GET', '/:c', () => {})
router.on('GET', '/text/hello', () => {})
router.on('GET', '/text/:e/test', () => {})

assert.deepEqual(router.find('GET', '/text/hellos/test').params, { e: 'hellos' })
```

PR #224, opened 2021-12-24
fix: parametric brother path extraction (#222)

bugagashenkj:
The problem was that after switching from static route to parametric brother route it was trying to restore the previous path by searching the last `/` symbol. It means that if I have two routes `/static-test` and `/static-:param` and if I try to find `/static-test1`, after failing with static route it will stitch to parametric route and pass `static-test1` as a param instead of `test1`.

I fixed it by creating a stack where I put the context (the current position in the path and the parameters already found) every time I meet a node that has a parametric brother. When I need to switch to a parametric brother route I just get a context from the top of the stack.

Another thing that I changed here is that I added a parametric brother route not only for static routes but for all routes.

### DIFF
--- a/custom_node.js
+++ b/custom_node.js
@@ -76,6 +76,7 @@ Node.prototype._saveParametricBrother = function () {
   let parametricBrother = this.parametricBrother
   for (const child of Object.values(this.children)) {
     if (child.prefix === ':') {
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

delvedor (approved):
LGTM

PR #225, opened 2021-12-25
change path for short static route bench

bugagashenkj:
When I tried to optimize the find method, I was confused. I made changes that were supposed to speed up the algorithm, but they only made it slower for short static routes. It turns out that `/` route is a super special case, cause you get the results immediately from the root node without using the searching algorithm. If you try to search the path that contains only one more symbol `/a`, you will get at least twice a worse result. IMHO, when we test the find method we want to see results for some average, but not for the extreme case.

### DIFF
--- a/bench.js
+++ b/bench.js
@@ -15,6 +15,7 @@ const FindMyWay = require('./')
 
 const findMyWay = new FindMyWay()
 findMyWay.on('GET', '/', () => true)
+findMyWay.on('GET', '/a', () => true)
 findMyWay.on('GET', '/user/:id', () => true)
 findMyWay.on('GET', '/user/:id/static', () => true)
 findMyWay.on('POST', '/user/:id', () => true)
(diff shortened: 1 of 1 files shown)


bench.js
```diff
@@ -14,7 +14,7 @@ const suite = Benchmark.Suite()
 const FindMyWay = require('./')
 
 const findMyWay = new FindMyWay()
-findMyWay.on('GET', '/', () => true)
```
mcollina, line 17: please retain this route

bench.js
```diff
@@ -74,7 +74,7 @@ suite
     constrained.lookup({ method: 'GET', url: '/versioned', headers: { 'accept-version': '2.x', host: 'fastify.io' } }, null)
   })
   .add('find static route', function () {
-    findMyWay.find('GET', '/', undefined)
+    findMyWay.find('GET', '/a', undefined)
```
mcollina, line 77: don't delete the `/` benchmark, add another one

mcollina (approved):
lgtm

PR #226, opened 2021-12-26
Optimize find method

bugagashenkj:
I refactored the find method. The main things that were changed:
- removed double path/prefix checking (in the find and findMatchingChild methods)
- removed unnecessary path splitting, I replaced it on just path pointer
- a bunch of other small optimizations

### DIFF
--- a/custom_node.js
+++ b/custom_node.js
@@ -135,16 +135,24 @@ Node.prototype.findByLabel = function (path) {
   return this.children[path[0]]
 }
 
-Node.prototype.findMatchingChild = function (derivedConstraints, path) {
-  var child = this.children[path[0]]
+Node.prototype.findMatchingChild = function (derivedConstraints, path, pathIndex) {
+  var child = this.children[path[pathIndex]]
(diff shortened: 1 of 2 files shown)


bugagashenkj:
depends on #224

bugagashenkj:
### Benchmark results:
**Before:**

Снимок экрана 2021-12-21 в 22 54 48

**After:**

image

bugagashenkj:
There is one thing that worries me. If someone can explain under what conditions this code can be executed, I will be very grateful. Because I think that this is an unreachable path of the code and tests also don't cover it.

https://github.com/delvedor/find-my-way/blob/9207236eb9a434293180c581c43bf7adcda1ee9d/index.js#L508

mcollina:
Good job!!!!

mcollina (approved):
lgtm

delvedor (commented):
Heya! The changes looks good, but I would like to take a little more time to review them! I'm currently on PTO, I'll take a look as soon as I'm back :)

delvedor (approved):
Great work!

PR #228, opened 2021-12-29
Added the parametric child property

bugagashenkj:
Added the new field parametricChild to the Node. It has the same logic as a wildcardChild property, but it is used for parametric nodes. It can optimize some calculations.


bugagashenkj:
depends on #226

bugagashenkj:
I close this since this logic is now contained in #230.

PR #229, opened 2021-12-30
Removed unnecessary constraints check

bugagashenkj:
It seems like it checks constraints twice: in the findMatchingChild function and in the find function.


bugagashenkj:
depends on #226

bugagashenkj:
### Benchmark results:
**Before (optimize-find-method, #226)**



**After:**

Снимок экрана 2021-12-30 в 14 44 45

bugagashenkj:
The biggest increase in performance is when we have constraints.

delvedor (changes requested):
Can you solve the conflicts first? It looks like some of this changes are already present in main, making this code harder to review.

delvedor (approved):
LGTM

PR #230, opened 2021-12-30
Fixed wildcard routes fallback

### LINKED ISSUE #227: Wildcard node should be checked for constraints
If a wildcard node does not pass constraints check we will return the default route even if there is another wildcard node that matches constraints.

There is a test for this case:
```js
test('Wildcard node with constraints', t => {
  t.plan(1)
  const findMyWay = FindMyWay({
    defaultRoute: (req, res) => {
    }
  })

    t.equal(params['*'], '/foo1/foo3')
  })

  findMyWay.on('GET', '/foo1/*', { host: 'fastify.io' }, (req, res, params) => {
  })

  findMyWay.on('GET', '/foo1/foo2', (req, res, params) => {
  })

  findMyWay.lookup(
    { method: 'GET', url: '/foo1/foo3', headers: {} },
    null
  )
})

```

bugagashenkj:
Fixed #227 
Depends on #229

### DIFF
--- a/test/issue-28.test.js
+++ b/test/issue-28.test.js
@@ -575,3 +575,29 @@ test('Nested wildcards with parametric and static - 8', t => {
     null
   )
 })
+
+test('Wildcard node with constraints', t => {
+  t.plan(1)
+  const findMyWay = FindMyWay({
+    defaultRoute: (req, res) => {
+      t.fail('we should not be here, the url is: ' + req.url)
+    }
+  })
+
+  findMyWay.on('GET', '*', (req, res, params) => {
+    t.equal(params['*'], '/foo1/foo3')
+  })
+
+  findMyWay.on('GET', '/foo1/*', { constraints: { host: 'fastify.io' } }, (req, res, params) => {
+    t.fail('we should not be here, the url is: ' + req.url)
+  })
+
+  findMyWay.on('GET', '/foo1/foo2', (req, res, params) => {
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

Eomm:
Did you run the benchmarks?
Could you share the results?

bugagashenkj:
### Benchmark results
**Main branch**

main

**Current branch**

Снимок экрана 2022-01-14 в 20 10 56

mcollina (approved):
lgtm

test/issue-28.test.js
```diff
@@ -575,3 +575,29 @@ test('Nested wildcards with parametric and static - 8', t => {
...
+    defaultRoute: (req, res) => {
+      t.fail('we should not be here, the url is: ' + req.url)
+    }
+  })
+
+  findMyWay.on('GET', '*', (req, res, params) => {
+    t.equal(params['*'], '/foo1/foo3')
+  })
+
+  findMyWay.on('GET', '/foo1/*', { constraints: { host: 'fastify.io' } }, (req, res, params) => {
```
Eomm, line 591: Could you add a test case that hit this route with constraint?
It is a valuable test that the defaultRoute is not hit in this case too

bugagashenkj: Done

mcollina:
@Eomm ping

Eomm (approved):

PR #231, opened 2021-12-31
Added maps for static routes

bugagashenkj:
The main idea is to add maps (Map<path, endpointNode>) for static routes. It should make them super fast, as we can get them for constant time without going through the tree.

With these changes, the long static bench speed increased more than 125% and it hasn't a lot of sub-nodes on its path that can split it. The more node you have, the slower your search is. But with the map, you will have always fast constant time results.

It's not the final version, I just want to hear your thoughts about it

See only the last commit.

### DIFF
--- a/custom_node.js
+++ b/custom_node.js
@@ -25,6 +25,7 @@ function Node (options) {
   this.regex = options.regex || null
   this.wildcardChild = null
   this.parametricChild = null
+  this.wildcardBrother = null
   this.parametricBrother = null
   this.constrainer = options.constrainer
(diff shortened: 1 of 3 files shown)


mcollina:
This seems a worthwhile improvement. However I'm not 100% sure how realistic it is to have such a long route.

delvedor (commented):
We tried this route in the past. If I recall correctly we didn't pursue it because in most of the cases, people use dynamic routes, and static routes are already quite fast.
Having two implementation internally might overcomplicate  things for a little benefit.

PR #233, opened 2022-01-05
Run benchmarks in separate threads

### LINKED ISSUE #232: Benchmark results depend on their order
There is something wrong happening with benchmarks. What I can say now is if you reorder your test cases you can get different results. I can't tell right now why it's happening, there can be myriads of reasons. I wanna start with the most extreme case: the first test case executes much faster than others. I was trying to test it on my laptop and my PC and got the same results. That's why I think that the "/" route executes much faster than others #225 (it's still a unique case, but the difference is not so dramatic). I was looking through the issues/PRs where people send their benchmarks and sometimes I saw the same picture (https://github.com/delvedor/find-my-way/pull/176#issuecomment-759957811). If you have the ability, please make some tests regarding the first test case. You can create multiple same tests and see if there are any differences depending on the test's place or just switch some tests and see if there is any difference.

bugagashenkj:
Fixed #232

### DIFF
--- a/bench-thread.js
+++ b/bench-thread.js
@@ -1,0 +1,44 @@
+'use strict'
+
+const { workerData, parentPort } = require('worker_threads')
+
+const Benchmark = require('benchmark')
+// The default number of samples for Benchmark seems to be low enough that it
+// can generate results with significant variance (~2%) for this benchmark
+// suite. This makes it sometimes a bit confusing to actually evaluate impact of
+// changes on performance. Setting the minimum of samples to 500 results in
+// significantly lower variance on my local setup for this tests suite, and
+// gives me higher confidence in benchmark results.
+Benchmark.options.minSamples = 500
+
+const suite = Benchmark.Suite()
+
+const FindMyWay = require('./')
+
+const findMyWay = new FindMyWay()
+
+const testingMethods = {
+  lookup: findMyWay.lookup,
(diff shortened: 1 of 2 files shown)


delvedor (commented):
Nice job! @mcollina any thoughts?

bugagashenkj:
### Benchmark results:
**Main branch:**

main

**Current branch:**

threads

**Current branch (shuffled):**

Снимок экрана 2022-01-12 в 23 12 59

mcollina (approved):
lgtm

mcollina:
This code is quite good!

bugagashenkj:
@delvedor I've found two duplicate benchmarks. Are they duplicates indeed or there should be some differences between them?

### **First:**
https://github.com/delvedor/find-my-way/blob/5beaacfcb6d1bb53663a90b5e7f0d5e9540e15ce/bench.js#L107
https://github.com/delvedor/find-my-way/blob/5beaacfcb6d1bb53663a90b5e7f0d5e9540e15ce/bench.js#L119

### **Second:**
https://github.com/delvedor/find-my-way/blob/5beaacfcb6d1bb53663a90b5e7f0d5e9540e15ce/bench.js#L116
https://github.com/delvedor/find-my-way/blob/5beaacfcb6d1bb53663a90b5e7f0d5e9540e15ce/bench.js#L122

PR #236, opened 2022-01-14
Fix issue in parameter resolution

### LINKED ISSUE #235: Issue in parameter resolution
Hi, 

I m using my find-my-way via Fastify and I m facing an issue parametrics url. 

I defined a route like that : `GET /images/:id.jpg`
When I make this request `GET /images/123456.jpg` it works as expected but when I make this one :  `GET /images/12346-fr.jpg`
It's does not work.

I think the issue is linked  to the support of multi params

fxalgrain:
Fix #235

### DIFF
--- a/index.js
+++ b/index.js
@@ -527,7 +527,19 @@ Router.prototype.find = function find (method, path, derivedConstraints) {
       } else {
         for (; paramEndIndex < pathLen; paramEndIndex++) {
           const charCode = path.charCodeAt(paramEndIndex)
-          if (charCode === 47 || charCode === 45 || charCode === 46) {
+          // Always break on /
+          if (charCode === 47) {
+            break
+          }
+          // Params start where next children begins
+          let isChildrenStartCharCode = false
+          for (let i = 0; i < node.childrenStartCodes.length; i++) {
+            if (charCode === node.childrenStartCodes[i]) {
+              isChildrenStartCharCode = true
+              break
+            }
+          }
+          if (isChildrenStartCharCode) {
             break
           }
         }
--- a/custom_node.js
+++ b/custom_node.js
@@ -20,7 +20,6 @@ function Node (options) {
   this.handlers = options.handlers || [] // unoptimized list of handler objects for which the fast matcher function will be compiled
   this.unconstrainedHandler = options.unconstrainedHandler || null // optimized reference to the handler that will match most of the time
   this.children = options.children || {}
-  this.numberOfChildren = Object.keys(this.children).length
   this.kind = options.kind || this.types.STATIC
   this.regex = options.regex || null
   this.wildcardChild = null
@@ -28,6 +27,8 @@ function Node (options) {
   this.constrainer = options.constrainer
   this.hasConstraints = options.hasConstraints || false
   this.constrainedHandlerStores = null
+
+  this._saveChildrenInfo()
 }
 
 Object.defineProperty(Node.prototype, 'types', {
@@ -64,9 +65,7 @@ Node.prototype.addChild = function (node) {
 
   this.children[label] = node
 
-  const nodeChildren = Object.values(this.children)
-  this.numberOfChildren = nodeChildren.length
-
+  this._saveChildrenInfo()
   this._saveParametricBrother()
 
   return this
@@ -93,6 +92,16 @@ Node.prototype._saveParametricBrother = function () {
   }
 }
 
+Node.prototype._saveChildrenInfo = function () {
+  const childrenKeys = Object.keys(this.children)
+  this.numberOfChildren = childrenKeys.length
+
+  this.childrenStartCodes = []
+  for (let i = 0; i < this.numberOfChildren; i++) {
+    this.childrenStartCodes.push(childrenKeys[i].charCodeAt(0))
+  }
+}
+
 Node.prototype.reset = function (prefix) {
   this.prefix = prefix
   this.children = {}
@@ -100,6 +109,7 @@ Node.prototype.reset = function (prefix) {
   this.unconstrainedHandler = null
   this.kind = this.types.STATIC
   this.numberOfChildren = 0
+  this.childrenStartCodes = []
   this.regex = null
   this.wildcardChild = null
   this.hasConstraints = false
--- a/test/path-params-match.test.js
+++ b/test/path-params-match.test.js
@@ -35,3 +35,18 @@ t.test('path params match', (t) => {
   t.equal(trailingSlashRet.handler, paramPath)
   t.same(trailingSlashRet.params, { pam: 'abcdef' })
 })
+
+t.test('should handles "-" in parameters', t => {
+  t.plan(2)
+  const findMyWay = FindMyWay()
+
+  findMyWay.on('GET', '/test/:id.jpg', () => {})
+
+  t.same(findMyWay.find('GET', '/test/123456.jpg').params, {
+    id: '123456'
+  })
+
+  t.same(findMyWay.find('GET', '/test/123456-78.jpg').params, {
+    id: '123456-78'
+  })
+})


index.js
```diff
@@ -525,9 +525,15 @@ Router.prototype.find = function find (method, path, derivedConstraints) {
         if (matchedParameter === null) return null
         paramEndIndex = pathIndex + matchedParameter[1].length
       } else {
+        const childrenStartCodes = node.numberOfChildren > 0 ? Object.keys(node.children).map(label => label.charCodeAt(0)) : []
```
mcollina, line 528: This looks slow, could you use a for(;;) loop insteaD?

index.js
```diff
@@ -525,9 +525,15 @@ Router.prototype.find = function find (method, path, derivedConstraints) {
...
         for (; paramEndIndex < pathLen; paramEndIndex++) {
           const charCode = path.charCodeAt(paramEndIndex)
-          if (charCode === 47 || charCode === 45 || charCode === 46) {
+          // Always break on /
+          if (charCode === 47) {
+            break
+          }
+          // Params start where next children begins
+          if (childrenStartCodes.includes(charCode)) {
```
mcollina, line 536: Includes is _really_ slow, could you use a for loop?

fxalgrain: Thanks for the review, I will comeback with improvements

fxalgrain:
@mcollina I refactored the code, taking into account your remarks. I hope it's good for you?

By the way could we expect this to be back port in the 4.x branch ?

bugagashenkj (changes requested):
It's shouldn't work this way. 
1. Your case is not a multiparametric route. There is a bug in the "_on" method. Your case is definitely just a one-parametric route.
2. If we want this case to be working, we need to define what characters can separate a parametric name from the static part. Now it's only "/", so it's unclear why "id.jpg" shouldn't be just a param's name. 
3. If you need this working right now, you can use regexp
```js
// parametric with regexp
router.on('GET', '/example/:file(^\\d+).png', () => {}))
```

index.js
```diff
@@ -525,9 +525,26 @@ Router.prototype.find = function find (method, path, derivedConstraints) {
         if (matchedParameter === null) return null
         paramEndIndex = pathIndex + matchedParameter[1].length
       } else {
+        const childrenStartCodes = []
+        for (let i = 0; i < node.numberOfChildren; i++) {
+          childrenStartCodes.push(node.childrenKeys[i].charCodeAt(0))
+        }
```
mcollina, line 531: Could this be pre-computed and attached to the node?

mcollina (commented):
good work!

mcollina:
@bugagashenkj I don't understand your point, how this should be fixed?

index.js
fxalgrain: I did it. I factorize the code  and fixed the reset method

bugagashenkj:
> fix: extract params (
At first, we need to fix that the '/test/:id.jpg' route setups as multiparametric, but it's obviously not.

bugagashenkj:
As a first thought, I would say that we need to start there:
https://github.com/delvedor/find-my-way/blob/5beaacfcb6d1bb53663a90b5e7f0d5e9540e15ce/index.js#L201

and fix it with that 

```js
      } else if (i < len && path.charCodeAt(i) !== 47 && path.charCodeAt(i + 1) === 58) {
```

mcollina:
@bugagashenkj could you include a test that would fail?

bugagashenkj:
> @bugagashenkj could you include a test that would fail?

I'm not sure, it's internal behavior. It defines it as a multiparametric route that is wrong and then escapes the second parameter because it's empty.

bugagashenkj:
@fxalgrain you can add one new route and see that it will break you previous one.
```js
t.test('should handles "-" in parameters', t => {
  t.plan(2)
  const findMyWay = FindMyWay()

  findMyWay.on('GET', '/test/:id.jpg', () => {})
  findMyWay.on('GET', '/test/:id-test', () => {})

  t.same(findMyWay.find('GET', '/test/123456.jpg').params, {
    id: '123456'
  })

  t.same(findMyWay.find('GET', '/test/123456-78.jpg').params, {
    id: '123456-78'
  })
})
```

fxalgrain:
I add the test and it fails. For information, I run the same test on the master of this repo, and I get the same result.

Test result with this PR: 
- ✅   `findMyWay.on('GET', '/test/:id.jpg', () => {})`
- 🔴 `findMyWay.on('GET', '/test/:id.jpg', () => {});  findMyWay.on('GET', '/test/:id-test', () => {});`

Test result before this PR:
- 🔴   `findMyWay.on('GET', '/test/:id.jpg', () => {})`
- 🔴 `findMyWay.on('GET', '/test/:id.jpg', () => {});  findMyWay.on('GET', '/test/:id-test', () => {});`

 ```
FAIL ​ test/path-params-match.test.js
 ✖ Cannot read property 'params' of null

  test/path-params-match.test.js                                        
  48 |   })                                                             
  49 |                                                                  
> 50 |   t.same(findMyWay.find('GET', '/test/123456-78.jpg').params, {  
     | -----------------------------------------------------^           
  51 |     id: '123456-78'                                              
  52 |   })                                                             
  53 | })                                                               

  test: should handles "-" in parameters
  stack: |
    Test.<anonymous> (test/path-params-match.test.js:50:54)
  type: TypeError
  tapCaught: testFunctionThrow
```

**Comment by bugagashenkj** [comment]
> @fxalgrain you can add one new route and see that it will break your previous one.
> 
> ```js
> t.test('should handles "-" in parameters', t => {
>   t.plan(2)
>   const findMyWay = FindMyWay()
> 
>   findMyWay.on('GET', '/test/:id.jpg', () => {})
>   findMyWay.on('GET', '/test/:id-test', () => {})
> 
>   t.same(findMyWay.find('GET', '/test/123456.jpg').params, {
>     id: '123456'
>   })
> 
>   t.same(findMyWay.find('GET', '/test/123456-78.jpg').params, {
>     id: '123456-78'
>   })
> })
> ```

We have this test already.
https://github.com/delvedor/find-my-way/blob/5beaacfcb6d1bb53663a90b5e7f0d5e9540e15ce/test/issue-17.test.js#L22

IMHO, we need to create clear documentation about special chars: ".", "/", "-"

**Comment by fxalgrain** [comment]
@bugagashenkj I agree about 

> IMHO, we need to create clear documentation about special chars: ".", "/", "-"

I think "-" should not be allowed in route definitions as it used as multi param splitter. And that's why if we add the `/test/:id-test` route, test fails. 

Maybe over my work, someone could work to make the separator char configurable.

**Comment by bugagashenkj** [comment]
@delvedor @mcollina  We use "." and "-" symbols to separate a non-regex parameter from another URL part. What do you think about ignoring these symbols? Should we support it? What the logic should be? It's an important question. 

I give you a non-obvious example:
```js
t.test('should handles "-" in parameters', t => {
  t.plan(1)
  const findMyWay = FindMyWay()

  findMyWay.on('GET', '/:param1-:param2', () => {})

  t.same(findMyWay.find('GET', '/p1-p2-p3').params, {
    param1: '???',
    param2: '???'
  })
})
```

mcollina:
> We use "." and "-" symbols to separate a non-regex parameter from another URL part. What do you think about ignoring these symbols? Should we support it? What the logic should be? It's an important question.

I don't know, I would not write such a route in the first place.

This is as much of a new feature as it can be. How would you like to handle them? Make a proposal.

bugagashenkj:
I don't create even a multiparametric route, but since it exists we need to support it somehow. I don't know the perfect solution here. If we make how this PR suggest, we will take a first matched symbol, so it will be

```js
t.test('should handles "-" in parameters', t => {
  t.plan(1)
  const findMyWay = FindMyWay()

  findMyWay.on('GET', '/:param1-:param2', () => {})

  t.same(findMyWay.find('GET', '/p1-p2-p3').params, {
    param1: 'p1',
    param2: 'p2-p3'
  })
})
```
But this solution has a bad behavior: if we add a new route it could break an old one.
Example:
This will work:
```js
t.test('should handles "-" in parameters', t => {
  t.plan(2)
  const findMyWay = FindMyWay()

  findMyWay.on('GET', '/test/:id.jpg', () => {})

  t.same(findMyWay.find('GET', '/test/123456.jpg').params, {
    id: '123456'
  })

  t.same(findMyWay.find('GET', '/test/123456-78.jpg').params, {
    id: '123456-78'
  })
})
```
But this won't:
```js
t.test('should handles "-" in parameters', t => {
  t.plan(2)
  const findMyWay = FindMyWay()

  findMyWay.on('GET', '/test/:id.jpg', () => {})
  findMyWay.on('GET', '/test/:id-test', () => {})

  t.same(findMyWay.find('GET', '/test/123456.jpg').params, {
    id: '123456'
  })

  t.same(findMyWay.find('GET', '/test/123456-78.jpg').params, {
    id: '123456-78'
  })
})
```

mcollina:
Thanks, I understand now.

fxalgrain:
> It's shouldn't work this way.
> 
> 1. Your case is not a multiparametric route. There is a bug in the "_on" method. Your case is definitely just a one-parametric route.
> 2. If we want this case to be working, we need to define what characters can separate a parametric name from the static part. Now it's only "/", so it's unclear why "id.jpg" shouldn't be just a param's name.
> 3. If you need this working right now, you can use regexp
> 
> ```js
> // parametric with regexp
> router.on('GET', '/example/:file(^\\d+).png', () => {}))
> ```
Thanks you @bugagashenkj, using a regex fix the issue in my case. I will notice in the issue that it's a valid workaround

PR #237, opened 2022-01-17
Fixed parameters parsing

### LINKED ISSUE #235: Issue in parameter resolution

bugagashenkj:
Fixed #235 #238

The main problem with multi-parametrical nodes is that we can't easily describe the rules of how should it be parsed. Thanks for @mcollina, he said that if I have an unclear case it should work as the [path-to-regexp](https://github.com/pillarjs/path-to-regexp) library does https://github.com/delvedor/find-my-way/issues/238#issuecomment-1021464532. The path-to-regexp converts the hole path to the regexp, so a multi-parametric path is also parsed as a regex. The main idea of this solution is to convert the multi-parametrical part of the path to the regexp. Seems like it's the only way to make it work properly.

**Important point: It doesn't convert the whole path to the regexp (as path-to-regexp does), only the part of the path between two '/' if it contains multiple params or param and static part.**

It replaces parameters with (.*?), regex parameters with themselves, the static part is also inserted into the regex.

Examples:
| Description | Multi-parametrical part of path  | Regexp | 
|---|---|---|
| Param with static part | /:param1-static/ | /^(.*?)-static$/ | 
| Multiple params | /:param1-:param2/ | /^(.\*?)-(.\*?)$/ | 
| Regexp param with param | /:param1(^\\w{3}).param2/ | /^(\w{3})\\.(.\*?)$/  | 

Depends on #243

### DIFF
--- a/custom_node.js
+++ b/custom_node.js
@@ -7,9 +7,7 @@ const types = {
   STATIC: 0,
   PARAM: 1,
   MATCH_ALL: 2,
-  REGEX: 3,
-  // It's used for a parameter, that is followed by another parameter in the same part
-  MULTI_PARAM: 4
+  REGEX: 3
 }
 
 function Node (options) {
@@ -158,7 +156,6 @@ Node.prototype.getChildByLabel = function (label, kind) {
       return this.wildcardChild
     case this.types.PARAM:
     case this.types.REGEX:
-    case this.types.MULTI_PARAM:
       return this.parametricChild
   }
 }
--- a/custom_node.js
+++ b/custom_node.js
@@ -46,7 +46,6 @@ Node.prototype.addChild = function (node) {
       break
     case this.types.PARAM:
     case this.types.REGEX:
-    case this.types.MULTI_PARAM:
       assert(this.parametricChild === null, 'There is already a parametric child')
       this.parametricChild = node
       break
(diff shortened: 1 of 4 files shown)


mcollina (commented):
Could you run the benchmarks before and after?

bugagashenkj:
**Main branch**

Снимок экрана 2022-03-31 в 14 00 51

**Current branch**

Current

bugagashenkj:
Yep, I can explain it. There is a regression for a multi-parameter route because previously two parametric nodes were used to parse paths like that `/:param1-:param2/`, which did not contain any regular expressions. It just parsed everything before and after the "-" as parameters. But there wasn't the right behavior. It worked only with simple paths. Now when it uses regex for this it obviously takes more time, but it works correctly in general.

About "multi-parametric route with regexp". It created two different nodes to parse paths like that `/:param1(^\w{3}).param2/`. The first is a param node with regex and the second is just a parametric node. Now it's just one regex parametric node. I guess that is the reason why it's working faster.

mcollina (approved):
lgtm

delvedor (approved):

PR #242, opened 2022-03-26
Fix double colon and parametric node collision

### LINKED ISSUE #241: Bug: Broken routing for aliased routes ending with `<...>/::something` & `<...>/:some_var`
**Versions tested:**
- 4.5.1 (Fastify dependency: `"find-my-way": "^4.5.0"`)
- 5.2.0 (latest)

*Both contain the bug, but the symptoms are different. The bug is more severe in 5.2.0.*

**TL;DR:** There appears to be a bug for pairs of routes, one ending with `/::something` and the other with `/:some_var`. The exact behavior depends on the `find-my-way` version (*see "Observed behavior" below*), but is incorrect in either case. *See "Failure MCVE" below.*

Prefixes other than `::` seem to work correctly (tested: `~`, `@`, `$`). *See "Working MCVE" below.*

---

**Failure MCVE:** (`index.mjs`)

```javascript
import http from 'http';
// npm install --save-prod find-my-way@4.5.1 or find-my-way@5.2.0
import FindMyWay from 'find-my-way';

const router = FindMyWay();

router.on('GET', '/:world_name/::articles', (req, res, params) => {
    res.end('matched /:world_name/::articles\n' + JSON.stringify(params, null, 2));
});
router.on('GET', '/:world_name/:article_name', (req, res, params) => {
    res.end('matched /:world_name/:article_name\n' + JSON.stringify(params, null, 2));
});

    router.lookup(req, res);
});

server.listen(3000, err => {
    if (err) throw err;
    console.log('Server listening on: http://localhost:3000');
});
```

**Observed behavior v4.5.1**: `GET /myworld/:articles` request matches the *second* (`/:world_name/:article_name`) route (see below).
**Observed behavior v5.2.0**: Both `GET /myworld/:articles` *and* `GET /myworld/somearticle` result in a 404 Not Found error. In other words, *neither* route works, as if they did not exist.
**Expected behavior**: The above request should match the *first* (`/:world_name/::articles`) route. Additionally (v5.2.0): The second route should still work as normal for `/myworld/somearticle`.
**Partial workaround**: No known workaround on my part for the *exact same* route, but changing `::articles` to a different prefix symbol (e.g. `~articles`) works in both 4.5.1 *and* 5.2.0.

`GET /myworld/:articles`:
```
matched /:world_name/:article_name
{
  "world_name": "myworld",
  "article_name": ":articles"
}
```

---

**Success MCVE:** (modified failure example, by changing `::articles` to `~articles` in the route)

```javascript
router.on('GET', '/:world_name/~articles', (req, res, params) => {
    res.end('matched /:world_name/~articles\n' + JSON.stringify(params, null, 2));
});
router.on('GET', '/:world_name/:article_name', (req, res, params) => {
    res.end('matched /:world_name/:article_name\n' + JSON.stringify(params, null, 2));
});
```

&hellip; (correctly) results in the following being matched for a `GET /myworld/~articles`:

```
matched /:world_name/~articles
{
  "world_name": "myworld"
}
```

bugagashenkj:
Fixed #241 

This is not the best solution, but I don't know how to fix it perfectly right now.

### DIFF
--- a/custom_node.js
+++ b/custom_node.js
@@ -19,8 +19,8 @@ function Node (options) {
   this.method = options.method // not used for logic, just for debugging and pretty printing
   this.handlers = options.handlers || [] // unoptimized list of handler objects for which the fast matcher function will be compiled
   this.unconstrainedHandler = options.unconstrainedHandler || null // optimized reference to the handler that will match most of the time
-  this.children = options.children || {}
-  this.numberOfChildren = Object.keys(this.children).length
+  this.staticChildren = options.staticChildren || {}
+  this.numberOfChildren = Object.keys(this.staticChildren).length
   this.kind = options.kind || this.types.STATIC
   this.regex = options.regex || null
   this.wildcardChild = null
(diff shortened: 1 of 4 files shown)


mcollina (approved):
LGTM

delvedor (changes requested):
The code looks ok.
Can you post a benchmark that compares this branch and main?

bugagashenkj:
**Main branch:**

main

**Current branch:**

Снимок экрана 2022-03-28 в 14 38 26

delvedor (approved):
LGTM

PR #243, opened 2022-03-28
Node insertion optimization

bugagashenkj:
As I mention here #242, it wasn't the best solution to this #241 issue. This is much better. I needed to optimize the insert method first. The main idea of optimization is that previously the insertion of each node started at the root of the tree. Now, setting a new node is based on the previous one. This allowed us to simplify the code a lot.

### DIFF
--- a/custom_node.js
+++ b/custom_node.js
@@ -146,6 +146,23 @@ Node.prototype.split = function (length) {
   return newChild
 }
 
+Node.prototype.getChildByLabel = function (label, kind) {
+  if (label.length === 0) {
+    return null
+  }
+
+  switch (kind) {
+    case this.types.STATIC:
+      return this.staticChildren[label]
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

RafaelGSS (commented):
Can you share the benchmarks?

delvedor (changes requested):
Nice! Can you share the benchmarks?

bugagashenkj:
**Main branch**

Снимок экрана 2022-03-29 в 12 13 51

**Current branch**

current

delvedor (approved):
LGTM

mcollina:
Could you rebase? Thanks!

bugagashenkj:
@mcollina #237 depended on this. Since you've merged it we, don't need to merge this anymore.

image

PR #248, opened 2022-04-12
Define insert method for each node type

bugagashenkj:
Just small refactoring. I split the insert method into three methods for each type of node (static/parametric/wildcard). IMHO it makes the code more self-explaining.


mcollina (approved):
lgtm

PR #249, opened 2022-04-12
Change ignore trailing slash behavior

bugagashenkj:
I changed ignoring trailing slash behavior. Instead of creating two duplicated routes (with slash and without), it just removes the trailing slash. Performance is the same, but it's much easier to maintain. It will affect the `routes` field. There will be only one trimmed route instead of two. The `route` field should be used only for debugging, so I don't know if it's a breaking change.

### DIFF
--- a/index.js
+++ b/index.js
@@ -115,21 +115,16 @@ Router.prototype.on = function on (method, path, opts, handler, store) {
     return
   }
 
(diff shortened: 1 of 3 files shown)


mcollina:
I don't see this as a breaking change.

PR #250, opened 2022-04-13
Create class for each node type

bugagashenkj:
I refactored the Node entity:
- Divided the Node into the three different classes: StaticNode, ParametricNode, WildcardNode; each of these classes has a different set of properties and methods
- Moved everything related to handlers and matching constraints to the separate entity: HandlerStorage

### DIFF
--- a/custom_node.js
+++ b/custom_node.js
@@ -1,358 +1,117 @@
 'use strict'
 
-const assert = require('assert')
-const deepEqual = require('fast-deep-equal')
+const HandlerStorage = require('./handler_storage')
 
-const types = {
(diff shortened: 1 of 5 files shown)


bugagashenkj:
**Main branch:**

main

**Current branch:**

current

mcollina:
Is the difference in the last benchmark in the script confirmed? Are we losing some perf?

bugagashenkj:
There is no reason for the last benchmark to be slower. I ran it again in the current branch.

Снимок экрана 2022-04-16 в 13 10 10

Benchmarks can have a significant margin of error between two runs.

Issue #254, opened 2022-04-22
`constraints` will be overrided when multiple router created


climba03003:
The current code override the `HandlerStorage.prototype.constrainer` to get the `constrainter` in router.
It will be override when two store created first before any route trying to lookup.

```js
const findMyWay = require("find-my-way");

const constraint = {
  name: "secret",
  storage: function () {
    let secrets = {};
    return {
      get: (secret) => {
        return secrets[secret] || null;
      },
      set: (secret, store) => {
        secrets[secret] = store;
      },
      del: (secret) => {
        delete secrets[secret];
      },
      empty: () => {
        secrets = {};
      },
    };
  },
  deriveConstraint: (req, ctx) => {
    return req.headers["x-secret"];
  },
  validate() {
    return true;
  },
};

let router1 = findMyWay({ constraints: { secret: constraint } });

// it override the constraints of router1
const router2 = findMyWay({ constraints: undefined });

// it can registered constraints
router1.on("GET", "/", { constraints: { secret: "alpha" } }, () => {});
// it cannot find the constraints secret
router1.find("GET", "/", { constraints: { secret: "alpha" } });
```

PR #255, opened 2022-04-22
fix: constrainer can be override by other instance

climba03003:
Fixes: #254 

We should not override the prototype again and again.

### DIFF
--- a/test/issue-254.test.js
+++ b/test/issue-254.test.js
@@ -7,8 +7,6 @@ const FindMyWay = require('..')
 test('Constraints should not be overrided when multiple router is created', t => {
   t.plan(1)
 
-  let count = 0
-
   const constraint = {
     name: 'secret',
     storage: function () {
@@ -30,10 +28,7 @@ test('Constraints should not be overrided when multiple router is created', t =>
   FindMyWay()
 
   router1.on('GET', '/', { constraints: { secret: 'alpha' } }, () => {})
-  count++
-
   router1.find('GET', '/', { constraints: { secret: 'alpha' } })
-  count++
 
-  t.equal(count, 2, 'constraints is not overrided')
+  t.pass('constraints is not overrided')
 })
--- a/test/issue-254.test.js
+++ b/test/issue-254.test.js
@@ -1,0 +1,39 @@
+'use strict'
+
+const t = require('tap')
+const test = t.test
+const FindMyWay = require('..')
+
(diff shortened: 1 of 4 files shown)


test/issue-254.test.js
```diff
@@ -0,0 +1,39 @@
...
+  let count = 0
+
+  const constraint = {
+    name: 'secret',
+    storage: function () {
+      let secrets = {}
+      return {
+        get: (secret) => { return secrets[secret] || null },
+        set: (secret, store) => { secrets[secret] = store },
+        del: (secret) => { delete secrets[secret] },
```
bugagashenkj, line 19: We don't use del and empty methods anymore.

test/issue-254.test.js
```diff
@@ -0,0 +1,39 @@
...
+  const router1 = FindMyWay({ constraints: { secret: constraint } })
+  FindMyWay()
+
+  router1.on('GET', '/', { constraints: { secret: 'alpha' } }, () => {})
+  count++
+
+  router1.find('GET', '/', { constraints: { secret: 'alpha' } })
+  count++
+
+  t.equal(count, 2, 'constraints is not overrided')
```
RafaelGSS, line 38: I don't get how this test guarantees the constraints are not overridden. Maybe you forgot to check if `router1.find` returns !== null?

bugagashenkj (commented):
I don't really like throwing handlers through node constructors. Please give me some time, I want to check if we can do something with it.

test/issue-254.test.js
climba03003: It must throw an error if the constraints is override. Because there is no secret constraints inside the Constrainer anymore.

climba03003: More breakdown.
1. adding `.on` is check by the `Router` options, so it will not be overrided.
2. using `.find` is the key point. The `constraints` function is used at this point, so it would throw an error if the `constraints` is override.

If the test catch the throwing error, that means the `constraints` is being overrided in `HandlerStorage`.

RafaelGSS: so, maybe is better to change `t.equal` to `t.pass`

bugagashenkj:
As I said I don't really want to have a constrainer in the node constructor. What do you think about this solution https://github.com/delvedor/find-my-way/pull/256

climba03003:
Closed in favour of #256

PR #256, opened 2022-04-22
Fix constrainer overwrite

### LINKED ISSUE #254: `constraints` will be overrided when multiple router created

bugagashenkj:
Fixed #254

### DIFF
--- a/handler_storage.js
+++ b/handler_storage.js
@@ -5,8 +5,6 @@ const deepEqual = require('fast-deep-equal')
 
 class HandlerStorage {
   constructor () {
-    this.constrainer = HandlerStorage.prototype.constrainer
-
     this.handlers = [] // unoptimized list of handler objects for which the fast matcher function will be compiled
     this.hasConstraints = false
     this.compiledHandler = null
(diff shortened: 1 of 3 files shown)


climba03003 (commented):
I will change the order of argument.
`constrainer` comes first then the other options.

It is more consistence in my eyes and `storage` usually not going to be removed / changed, but options depends on it will.

bugagashenkj:
I changed the order

climba03003 (approved):
Nice

RafaelGSS (approved):
LGTM.

mcollina (approved):
lgtm

PR #267, opened 2022-04-29
Normalize benchmark paths

ivan-tymoshenko:
Benchmark refactoring:
- Updated benchmarks paths. All benchmark test routes have different paths and it takes significant time to iterate a path char by char. So some test cases can look really slow, not for the reason we are trying to test, but just because they have lost static prefix. I made paths the same where it makes sense, so it's easier to compare benchmarks between themselves.
- Removed find method benchmarks. Almost all find method benchmarks are duplicating lookup benchmarks. They have some sense because lookup adds some overhead by deriving constraints and calling a handler. But since we use an empty handler and almost empty constraint this overhead is really small and constant. From my experience find method benchmarks don't give any new information, after running lookup benchmarks.
- Added some new benchmarks.
- Aligned test names.

Снимок экрана 2022-04-29 в 20 09 13

### DIFF
--- a/bench-thread.js
+++ b/bench-thread.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const { workerData, parentPort } = require('worker_threads')
+const { workerData: benchmark, parentPort } = require('worker_threads')
 
 const Benchmark = require('benchmark')
 // The default number of samples for Benchmark seems to be low enough that it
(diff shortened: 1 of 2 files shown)


RafaelGSS (approved):
Great. LGTM

mcollina (approved):
lgtm

PR #268, opened 2022-04-30
Add comparison of benchmarks results on different branches

ivan-tymoshenko:
Added two new commands for benchmarking.
- `bench:cmp` - allows to choose two branches, run benchmarks and compare results.
- `bench:cmp:ci` - runs `bench:cmp` with default params current branch as a "new feature" branch and main branch.

If the difference between benchmarks is more or less than the threshold = 5%, it will emphasize the benchmark result with green or red color. 

selecting branches:

Снимок экрана 2022-04-30 в 21 44 30

`bench:cmp:ci` output example:

Снимок экрана 2022-04-30 в 21 39 41

### DIFF
--- a/bench-cmp.js
+++ b/bench-cmp.js
@@ -1,0 +1,114 @@
+'use strict'
+
+const { spawn } = require('child_process')
+
+const chalk = require('chalk')
+const inquirer = require('inquirer')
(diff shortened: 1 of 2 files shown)


mcollina (approved):
amazing! LGTM!

Issue #240, opened 2022-03-03
Unexpected behavior on double slash in path


code-ape:
## Description

Today, find-my-way handles empty variables in a way that feels very unintuitive. For example, the route pattern of `/a/:b/:c/d` matches `/a///d`. This is very confusing and also causes a "gap" with fastify and fastify-swagger (see last section for details).

I'm unsure if this is expected behavior or a bug? If it is expected, would it be crazy to offer a configuration option to collapse repeated `/`, as most HTTP servers do?

Let me know and thanks for taking the time to reply 🙏

## Example

```js
'use strict'

/**
 * Prints out result of test based on the issuer submitters expectations versus current results:
 *
 * [PASS|FAIL] METHOD URL,  expected: FUNCTION,   got: FUNCTION
 *
 * To test that no match is found pass `null` for `expectedFunction`
 */
function testPath(method, path, expectedFunction) {
  const result = findMyWay.find(method, path)
  // Get names of found function and expected function
  const resultFnName = result !== null ? result.handler.name : 'NONE'
  const expectedFnName = expectedFunction !== null ? expectedFunction.name : 'NONE'
  // Get pass/fail text based on match
  const resultString = resultFnName === expectedFnName ? '[PASS]' : '[FAIL]'
  // Print result
  console.log(`${resultString} ${method} ${path}, \texpected: ${expectedFnName}, \tgot: ${resultFnName}`)
}

const FindMyWay = require('find-my-way')
const findMyWay = FindMyWay({ ignoreTrailingSlash: true })

const handler1 = () => true
findMyWay.on('GET', '/a/:b/:c/d', handler1)

testPath('GET', '/a/b/c/d', handler1)
// Result: [PASS] GET /a/b/c/d,    expected: handler1,     got: handler1

testPath('GET', '/a//c/d', null)
// Result: [FAIL] GET /a//c/d,     expected: NONE,         got: handler1

testPath('GET', '/a///d', null)
// Result: [FAIL] GET /a///d,      expected: NONE,         got: handler1

testPath('GET', '/a//d', null)
// Result: [PASS] GET /a//d,       expected: NONE,         got: NONE
```

## fastify-swagger complication

This ticket comes from a strange issue where there was a "gap" between fastify-swagger and Fastify. In essence, if a route parameter was an empty string (see above for details) then fastify-swagger would fail to match but fastify would. This meant that validation was bypassed but the handler still got invoked.

mcollina:
[...]

PR #270, opened 2022-05-01
Adding feature ignoreDuplicateSlashes with tests

code-ape:
## Overview
This is a PR for the feature discussed in https://github.com/delvedor/find-my-way/issues/240 which relates to expected behavior for duplicate slashes in path. It was stated this would be a welcome PR and @mcollina directed me to PR #50 for reference of implementing a similar feature.

## Related Issues & PRs

1. #240
2. #50

## Changes
1. Add support for config option `ignoreDuplicateSlashes` which will condense any continuous occurrence of `/` in the path into a single `/`.
2. Add tests for `ignoreDuplicateSlashes` following pattens of existing tests for `ignoreTrailingSlash`.
3. Basic documentation of this feature added to `README.md`.


index.js
```diff
@@ -564,6 +580,10 @@ function trimLastSlash (path) {
   return path
 }
 
+function removeDuplicateSlashes (path) {
+  return path.replace(/\/\/+/g, '/')
```
ivan-tymoshenko, line 584: It's not a good idea to use regex matching in a find method. It's a hot path. You can run benchmarks to see the performance drop.

ivan-tymoshenko (changes requested):
Hi, thanks for your contribution. I guess the main idea was to not match the empty params, but not to remove duplicate slashes. I mean if someone decided to add a route with two slashes in a row we should match it.

Try to check if a parameter length is not equal to 0, how we check the maximum parameter length and see if it works. https://github.com/delvedor/find-my-way/blob/2608f2baf0ff6be19f2616bae8682ddfbd8f675b/index.js#L446

code-ape:
Hi @ivan-tymoshenko, thanks for the quick reply 😄

I think you may have misunderstood #240. I filed it a couple months ago and the desired functionality was to remove duplicate sequential `/`. This is common behavior for web servers, for example this works: https://github.com/delvedor//////find-my-way

You can check the code example which has paths like `/a///d` as test cases. I was opening this PR based on this comment which indicted it would be of interest: https://github.com/delvedor/find-my-way/issues/240#issuecomment-1058449629

And thanks for the callout on regex! I will replace that with a proper loop shortly for better performance 🚀

code-ape:
I did a hand coding of the method for removing duplicate slashes and a benchmark to compare it to Regex. All work for that has been pushed to this branch. Both seem to perform well and the regex appears to do better in more "real world" examples. Below is a run from my M1 MacBook Pro.

Slowest performing scenario seems to be ~150ns / function call or ~500 CPU clock cycles for a 40 character string, which comes to ~10 CPU cycles per character. Also note that "no-op" scenarios where there are no duplicate slashes (which is presumably the majority of paths that would be processed in the real world) appears to only ~3 CPU cycles per character! 😄

```
$ node ./bench-duplicate.js
[/a/b]
removeDuplicateSlashes(0)      x 96,554,737 ops/sec ±0.21% (296 runs sampled)
removeDuplicateSlashesRegex(0) x 39,600,222 ops/sec ±0.10% (295 runs sampled)

[/api/v1/myNamespace/myObject/action1] 
removeDuplicateSlashes(1)      x 12,750,900 ops/sec ±0.41% (292 runs sampled)
removeDuplicateSlashesRegex(1) x 25,132,656 ops/sec ±0.29% (295 runs sampled)

[/a//b]
removeDuplicateSlashes(2)      x 35,200,687 ops/sec ±0.21% (297 runs sampled)
removeDuplicateSlashesRegex(2) x 18,131,065 ops/sec ±0.49% (289 runs sampled)

[/api/v1/myNamespace/myObject//action1]
removeDuplicateSlashes(3)      x 9,241,756 ops/sec ±0.31% (293 runs sampled)
removeDuplicateSlashesRegex(3) x 15,398,355 ops/sec ±0.36% (290 runs sampled)

[/a///////////////////////b]
removeDuplicateSlashes(4)      x 12,526,865 ops/sec ±0.32% (292 runs sampled)
removeDuplicateSlashesRegex(4) x 15,384,914 ops/sec ±0.20% (295 runs sampled)

[/api/v1/myNamespace/myObject///////////////////////action1]
removeDuplicateSlashes(5)      x 5,976,074 ops/sec ±0.32% (293 runs sampled)
removeDuplicateSlashesRegex(5) x 13,493,742 ops/sec ±0.21% (291 runs sampled)

[/api//v1//myNamespace//myObject//action1]
removeDuplicateSlashes(6)      x 6,328,285 ops/sec ±0.23% (294 runs sampled)
removeDuplicateSlashesRegex(6) x 6,712,087 ops/sec ±0.18% (295 runs sampled)
```

ivan-tymoshenko:
Hmm, it looks like regex in that case works really faster. Can you provide `npm run bench:cmp:ci` command results?

ivan-tymoshenko:
Another issue I'm worried about is that we shouldn't modify a wildcard part of the path. You can see that we return a wildcard param without trimming the last slash.

lib/duplicate-slashes.js
```diff
@@ -0,0 +1,37 @@
+function removeDuplicateSlashes (path) {
```
ivan-tymoshenko, line 1: Can you remove this, since you discovered that the regexp version is working faster?

bench-duplicate.js
```diff
@@ -0,0 +1,41 @@
+'use strict'
```
ivan-tymoshenko, line 1: We don't need these benchmarks anymore.

ivan-tymoshenko (commented):
@code-ape Hi, do you need any help? Can you remove non-regexp implementation and provide benchmark results?

ivan-tymoshenko:
Снимок экрана 2022-05-24 в 12 59 08

mcollina (approved):
lgtm

PR #273, opened 2022-05-05
Refactored handler storage

ivan-tymoshenko:
Refactored and optimized handler storage.

### DIFF
--- a/handler_storage.js
+++ b/handler_storage.js
@@ -1,101 +1,99 @@
 'use strict'
 
-const assert = require('assert')
 const deepEqual = require('fast-deep-equal')
 
 class HandlerStorage {
   constructor () {
-    this.handlers = [] // unoptimized list of handler objects for which the fast matcher function will be compiled
-    this.hasConstraints = false
-    this.compiledHandler = null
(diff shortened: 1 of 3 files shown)


ivan-tymoshenko:
There are only changes in the constrained routes in the benchmarks. I measured it a few times. All other changes are just benchmark mistakes.

Снимок экрана 2022-05-05 в 21 19 53

Снимок экрана 2022-05-05 в 21 23 52

mcollina (approved):
lgtm

handler_storage.js
```diff
@@ -1,101 +1,99 @@
...
+          // always check the version constraint first as it is the most selective
+          this.constraints.unshift(constraint)
+        } else {
+          this.constraints.push(constraint)
+        }
+      }
     }
 
-    if (this.hasConstraints && this.handlers.length > 32) {
+    if (this.constrainedHandlers.length >= 32) {
```
mcollina, line 60: why this has became a `>=` check instead of a `>` check?

ivan-tymoshenko: Because before we made this check after inserting a handler into the handler's array and now we check it before inserting it.

PR #275, opened 2022-05-08
Fixed must match constraints check

ivan-tymoshenko:
The is a bug with constraints with the mustMatchWhenDerived option.

How it should work. If you have a constraint with the mustMatchWhenDerived option in your request you can match only the routes that have and match this constraint.

How it works. If you have a constraint with the mustMatchWhenDerived option in your request you can match the routes that have and match this constraint or other routes that have any other different constraints.

There are some tests that test wrong behavior:
https://github.com/delvedor/find-my-way/blob/a59117d0daf351429919d586e5edea4ef6d2a14b/test/constraints.test.js#L62
https://github.com/delvedor/find-my-way/blob/a59117d0daf351429919d586e5edea4ef6d2a14b/test/constraints.test.js#L75
https://github.com/delvedor/find-my-way/blob/a59117d0daf351429919d586e5edea4ef6d2a14b/test/constraints.test.js#L88

Benchmarks:

Снимок экрана 2022-05-08 в 16 31 43

### DIFF
--- a/handler_storage.js
+++ b/handler_storage.js
@@ -6,21 +6,12 @@ class HandlerStorage {
   constructor () {
     this.unconstrainedHandler = null // optimized reference to the handler that will match most of the time
     this.constraints = []
-    this.constrainedHandlers = [] // unoptimized list of handler objects for which the fast matcher function will be compiled
+    this.handlers = [] // unoptimized list of handler objects for which the fast matcher function will be compiled
     this.constrainedHandlerStores = null
   }
 
   hasHandler (constraints) {
-    if (Object.keys(constraints).length === 0) {
-      return this.unconstrainedHandler !== null
-    }
-
-    for (const handler of this.constrainedHandlers) {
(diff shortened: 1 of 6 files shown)


mcollina:
could you fix the performance regression?

ivan-tymoshenko:
The long static route doesn't have any regression as well as wildcard node doesn't have any speed up. It's a measuring error. The unconstraint route has performance regression. I had to remove some optimizations that I made in the previous PR. https://github.com/delvedor/find-my-way/pull/273. I can't do anything with it.

mcollina (approved):
lgtm

ivan-tymoshenko:
I will explain the problem one more time. When we decode a path with the decodeURI function, it decodes some chars and shrinks the path, so we can't tell where a param/wildcard part starts anymore.

ivan-tymoshenko:
@mcollina Can we merge this PR?

mcollina (approved):
lgtm

PR #278, opened 2022-05-10
Change lookup result type to any

### LINKED ISSUE #274: Wait for async handler to resolve
Would it be possible to wait for the async handler to resolve in the following code? How could we know when the lookup is done

```ts
fmw.on('GET', '/', async (req, res, params) => {
    // Slow handler that could take some time
});

const data = await before(req, res);

// `await` awaits the async handler
await fmw.lookup(req, res);

await after(data);
```

ivan-tymoshenko:
Fixed #274

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -55,7 +55,7 @@ declare namespace Router {
     res: Res<V>,
     params: { [k: string]: string | undefined },
     store: any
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #279, opened 2022-05-11
Add addConstraintStrategies method

### LINKED ISSUE #272: Add constraint outside the constructor
Currently, constraints can only be added when creating the FindMyWay instance. It would be very useful to be able to add more constraints later via a method like ``addConstraint``. With regard to Fastify, this would allow plugins to add constraints without the developer having to specify the constraint itself when creating the Fastify instance.

```js

router.addConstraint(...)
```

Is this even implementable and if so is it desirable to have such functionality?

ivan-tymoshenko:
Close #272

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,9 @@ Constraints can be combined, and route handlers will only match if __all__ of th
 <a name="custom-constraint-strategies"></a>
 ### Custom Constraint Strategies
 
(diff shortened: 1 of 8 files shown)


ivan-tymoshenko:
@matthyk Does it work for you?

mcollina (approved):
lgtm

ivan-tymoshenko:
Adjustment PR: https://github.com/delvedor/find-my-way/pull/287

PR #282, opened 2022-05-13
Fix parameters decoding

ivan-tymoshenko:
Fix #234 and a bunch of optimization to make it work as fast as possible.

PR #286, opened 2022-05-17
Remove decodeUriParameters option

ivan-tymoshenko:
Depends on #282 

After merging #282, there would be no sense in this option. The `decodeUriParameters` option is used to optimize decoding non-Latin chars in path params. After #282 we won't parse params from the raw path but will parse them from the decoded path by the decodeURI function. The only chars that we will need to decode in params are `# $ & + , / : ; = ? @ %`. We can do it natively in the find-my-may in the most optimized way https://github.com/delvedor/find-my-way/blob/d83459668afad494ceb82fe7244b20ceb8238a6e/lib/url-sanitizer.js#L69.

If you didn't add any side effects to this function, it doesn't change anything for you except increasing performance.

Fastify doesn't use the `decodeUriParameters` option.


mcollina (approved):
lgtm

mcollina (approved):
lgtm

PR #291, opened 2022-05-25
Add multiple regexp parametric children

### LINKED ISSUE #285: Parametric regex match fails with similar routes
Using two routes which start with a regex param, the second route which should match `/b/static/` doesn't.

```javascript
findMyWay.on('GET', '/:a(a)', () => {})
findMyWay.on('GET', '/:b(b)/static', () => {})
```

Here is a complete test:

```javascript
'use strict'

const t = require('tap')
const test = t.test
const FindMyWay = require('../')

test('Parametric regex match with similar routes', (t) => {
  t.plan(1)
  const findMyWay = FindMyWay()

  findMyWay.on('GET', '/:a(a)', () => {})
  findMyWay.on('GET', '/:b(b)/static', () => {})

  t.same(findMyWay.find('GET', '/b/static', {}).params, { 'b': 'b' })
})
```

ivan-tymoshenko:
Fixed #285

### DIFF
--- a/custom_node.js
+++ b/custom_node.js
@@ -56,18 +56,21 @@ class StaticNode extends ParentNode {
     super()
     this.prefix = prefix
     this.wildcardChild = null
-    this.parametricChild = null
+    this.parametricChildren = []
     this.kind = NODE_TYPES.STATIC
     this._compilePrefixMatch()
   }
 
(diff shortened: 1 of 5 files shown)


mcollina (approved):
lgtm

PR #292, opened 2022-05-26
Compare regexp by regexp.source

ivan-tymoshenko:


RafaelGSS (commented):
What does it solve/improve?

ivan-tymoshenko:
If there is a two parametric nodes in a tree that has the same regex source we can consider them as one node, instead of two different. It will not fix any bug.

mcollina (approved):
lgtm

mcollina:
I would prefer to have a test for this change. Why was the original code wrong?

ivan-tymoshenko:
It was incorrect, but there wasn't a bug. It wasn't optimized. It was wrong, because regex is an object and comparing object like that would always give false. But it will work in both cases. If it's unoptimized it will create two different branches for two nodes with the same regex.source. If it optimized it will create one branch.

To test things like that we need to test internal tree stucture, which is not a part of api. Don't know if we should do that.

PR #293, opened 2022-06-01
fix: remove max param length on wildcard

climba03003:
Fixes https://github.com/fastify/fastify/issues/3957

It is very common that people expect `wildcard` route as a fallback / catch-all route in their application.
Since, it will match all the thing include each path segment. 
Restricting the length of `wildcard` like the other `param / regexp` seems unnecessary and out of user expectation.

I think this behavior is bring from https://github.com/delvedor/find-my-way/pull/237


ivan-tymoshenko (approved):
lgtm

PR #299, opened 2022-08-23
fix matcher function in off method and fix regex tests

matthyk:
The `off` method had some bugs.

```js

router.on('GET', '/test', {}, (req, res) => {})
router.on('GET', '/test', { constraints: { host: 'example.com' } }, (req, res) => {})

router.off('GET', '/test', { host: 'example.com' })

router.prettyPrint() // is empty
```

or

```js

router.on('GET', '/test', {}, (req, res) => {})
router.on('GET', '/test', { constraints: { host: 'example.com' } }, (req, res) => {})

router.off('GET', '/test', { host: 'test.de' }) // only removes the route without constraint
```

The `off` method now implements the following logic:

1. `router.off` is called without constraints: Any route matching the provided method and path are removed regardless of the constraints
2. `router.off` is called with constraints: Only the route with the exact same constraints will be removed. Routes without constraints are not removed.  

In the regex.test.js file, the `off` method was called incorrectly, resulting in incorrect tests. This has been fixed as well.

### DIFF
--- a/test/methods.test.js
+++ b/test/methods.test.js
@@ -775,7 +775,7 @@ test('off removes all routes without checking constraints if no constraints are 
 })
 
 test('off removes only constrainted routes if constraints are specified', t => {
-  t.plan(1)
+  t.plan(2)
 
   const findMyWay = FindMyWay()
 
@@ -785,6 +785,7 @@ test('off removes only constrainted routes if constraints are specified', t => {
   findMyWay.off('GET', '/test', { host: 'example.com' })
 
   t.equal(findMyWay.routes.length, 1)
+  t.notOk(findMyWay.routes[0].opts.constraints)
 })
 
 test('off removes no routes if provided constraints does not match any registered route', t => {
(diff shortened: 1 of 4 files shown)


ivan-tymoshenko (approved):
LGTM

mcollina (approved):
lgtm

test/methods.test.js
```diff
@@ -760,3 +760,43 @@ test('register all known HTTP methods', t => {
...
+
+test('off removes all routes without checking constraints if no constraints are specified', t => {
+  t.plan(1)
+
+  const findMyWay = FindMyWay()
+
+  findMyWay.on('GET', '/test', {}, (req, res) => {})
+  findMyWay.on('GET', '/test', { constraints: { host: 'example.com' } }, (req, res) => {})
+
+  findMyWay.off('GET', '/test')
```
Eomm, line 772: If I want to remove the first `/test` route without constraints, 
will the following code work?

```
findMyWay.off('GET', '/test', {})
```

matthyk: No but you could delete both route with 
```js
findMyWay.off('GET', '/test')
``` 
But you are right that there is currently no way to delete only the route without constraint. But that should be possible. But then it is no longer possible to delete multiple routes that differ only in the constraint with one call. One possibility would be to introduce an option like `deleteAll`, which would then take care of deleting all routes that differ only by the constraint.

ivan-tymoshenko: @matthyk If the absence of an argument will delete all routes, and an empty object will only delete a route without constraints, will it be convenient?

matthyk: @ivan-tymoshenko Absolutely not 😄 I'll look at it again tomorrow morning

ivan-tymoshenko: @matthyk why not? It sounds pretty logical to me. If you pass the constraint argument, you want to delete only the route that match the constrains. Route with no constraints will match with empty object. If you want to delete the route with all it's constrained variations you will skip the constraint argument.

Eomm: From my point of view, given the API `findMyWay.off(method, path, constraints)`

- if the constraint is `=== undefined`, it is like calling `findMyWay.off(method, path)`
- if the constraint is falsy value (except `undefined`), it should match the route without constraints (as preference, I thing the empty object `{}` works as well)

matthyk: Sorry for the late response. @ivan-tymoshenko and @Eomm you have convinced me 👍 
what do you think about doing it like this?

- if the constraint is any falsy value (including `undefined`) it's like calling `findMyWay.off(method, path)` and all routes, regardless of the constraints, are deleted.
- if the constraint is an empty object only the route without constraints is deleted.

This is somewhat different in the handling of falsy values than the suggestion of @Eomm

ivan-tymoshenko: - if you want to delete only an unconstrained route, you should pass the empty object. This is the only way to do it.
- if you don't pass any parameter, except undefined, because it literally means nothing, all routes will be deleted.
- if you pass something other than an object, it should be your problem. We shouldn't try to guess what you want to do.

matthyk: So you suggest:

- findMyWay.off(method, path)  or findMyWay.off(method, path, undefined) &rarr; all routes will be deleted
- findMyWay.off(method, path, {}) &rarr; only the unconstrainted route will be deleted
- the third argument is neither undefined nor of type object &rarr; we will throw an error

ivan-tymoshenko: yes.

Eomm (approved):

PR #300, opened 2022-09-05
feat: replace querystring with fast-querystring

anonrig:
Solves issue: https://github.com/fastify/fastify/issues/4252

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ const router = require('find-my-way')({
 })
 ```
 
-The default query string parser that find-my-way uses is the Node.js's core querystring module. You can change this default setting by passing the option querystringParser and use a custom one, such as [qs](https://www.npmjs.com/package/qs).
+The default query string parser that find-my-way uses is [fast-querystring](https://www.npmjs.com/package/fast-querystring) module. You can change this default setting by passing the option querystringParser and use a custom one, such as [qs](https://www.npmjs.com/package/qs).
 
 ```js
 const qs = require('qs')
--- a/index.js
+++ b/index.js
@@ -27,7 +27,7 @@
 
 const assert = require('assert')
 const http = require('http')
-const querystring = require('querystring')
+const querystring = require('fast-querystring')
 const isRegexSafe = require('safe-regex2')
 const deepEqual = require('fast-deep-equal')
 const { flattenNode, compressFlattenedNode, prettyPrintFlattenedNode, prettyPrintRoutesArray } = require('./lib/pretty-print')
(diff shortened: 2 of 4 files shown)


anonrig:
@delvedor I added all of the node.js querystring tests and released a new version (0.3.0) with the necessary changes to replicate the behavior of the `node:querystring` package.

ivan-tymoshenko (approved):
LGTM! Update the doc, please.

anonrig:
@ivan-tymoshenko I just updated `fast-querystring` module to 0.3.1 which comes with **~7%** faster parsing and updated the README.

anonrig:
Updated to 0.4.0 with lots of great improvements.

Fdawgs:
This repo supports node >=14, but `fast-querystring` uses `replaceAll()`, which requires Node >=15.

mcollina:
This will likely land in a major version, ideally shipping February/March 2023, in which we will drop support for Node v14 and v16. 

(v16 LTS ends Sept 2023 due to OpenSSL).

anonrig:
I might replace `replaceAll` with `replace` if that is a deal breaker. The performance difference is acceptable since even without it, the package is quite fast. @Fdawgs

anonrig:
FYI: Updated this pull-request to 0.7.0 (40% faster than querystring atm), dropped replaceAll in favor of supporting Node 14.

mcollina (approved):
lgtm, I think this would be a significant benefit for the project.

RafaelGSS (approved):
LGTM. Good job!

anonrig:
I released and updated the pull request to use 1.0.0 version. I think `fast-querystring` is stable.

delvedor (approved):
LGTM

mcollina:
@delvedor should I cut a major?

ivan-tymoshenko:
@mcollina @delvedor Can you explain to me why we should land it as a major? It's not a breaking change. It's not even a new feature. It might have a good querystring performance parsing upgrade, but it will not make a significant performance difference along with routing. And we have the 7.1.0 version now.  Why can't we land it as a 7.2.0?

mcollina:
I'm ok with a minor too.

anonrig:
If we are ok with this pull request, can we merge & release @mcollina @delvedor?

delvedor:
> Can you explain to me why we should land it as a major? 

The main reason is that this module has an average of 3.6 millions downloads per months, and you don't want to risk it :D
If we are confident there are no hidden breaking changes, we can do a minor.

mcollina:
I'll try a minor and in case revert

PR #246, opened 2022-04-09
Optimize constraints' strategies

bugagashenkj:
I optimized the version constraint strategy and removed empty and delete methods from the strategy interface. I don't know why we need them.

PR #301, opened 2022-09-07
make del and empty optional

matthyk:
With #246 the methods `del` and `empty` from the constraint storage object were removed from the documentation, since these two methods are not used anyway. However, they should then also be removed in the types.

I made it optional for now, since otherwise it would be a major change? In the next major release you could then remove them completely.


ivan-tymoshenko (approved):

mcollina (approved):
lgtm

PR #305, opened 2022-09-25
feat: add async custom constraints support

### LINKED ISSUE #303: async deriveConstraint support
I have through about it a long time ago and tried hard to prevent this usage in my application.

The use-case I have is derive the route based on `JWT` token.
I contain all the required `permission`, so the `router` can reach their dedicated `handler` based on `JWT` token.

In a relative small scale project, it works absolutely fine and I am happy with the result.
But somehow when the project growth and `JWT` token will extended to a very long string.
It will exceed the limit of `header` from `node`, `reverse proxy`, etc. and increasing the `header` limit endlessly is not a good solution.

Somehow, I need to `fetch` the `permission` inside database to prevent the extend of `JWT` size.

ivan-tymoshenko:
Close #303

### DIFF
--- a/lib/constrainer.js
+++ b/lib/constrainer.js
@@ -65,7 +65,7 @@ class Constrainer {
       return constraints
     }
 
-    this.deriveAsyncConstraints(constraints || {}, req, ctx, done)
+    this.deriveAsyncConstraints(constraints, req, ctx, done)
   }
 
   deriveSyncConstraints (req, ctx) {
@@ -116,6 +116,12 @@ class Constrainer {
   deriveAsyncConstraints (constraints, req, ctx, done) {
     let asyncConstraintsCount = this.asyncStrategiesInUse.size
 
+    if (asyncConstraintsCount === 0) {
+      done(constraints)
+      return
+    }
+
+    constraints = constraints || {}
     for (const key of this.asyncStrategiesInUse) {
       const strategy = this.strategies[key]
       strategy.deriveConstraint(req, ctx, constraintValue => {
--- a/lib/constrainer.js
+++ b/lib/constrainer.js
@@ -22,12 +22,16 @@ class Constrainer {
     }
   }
 
+  isStrategyUsed (strategyName) {
+    return this.strategiesInUse.has(strategyName) ||
+      this.asyncStrategiesInUse.has(strategyName)
+  }
+
   hasConstraintStrategy (strategyName) {
     const customConstraintStrategy = this.strategies[strategyName]
     if (customConstraintStrategy !== undefined) {
       return customConstraintStrategy.isCustom ||
-        this.strategiesInUse.has(strategyName) ||
-        this.asyncStrategiesInUse.has(strategyName)
+        this.isStrategyUsed(strategyName)
     }
     return false
   }
@@ -41,7 +45,7 @@ class Constrainer {
       throw new Error(`There already exists a custom constraint with the name ${strategy.name}.`)
     }
 
-    if (this.strategiesInUse.has(strategy.name) || this.asyncStrategiesInUse.has(strategy.name)) {
+    if (this.isStrategyUsed(strategy.name)) {
       throw new Error(`There already exists a route with ${strategy.name} constraint.`)
     }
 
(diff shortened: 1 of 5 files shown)


mcollina:
Can you add docs?

ivan-tymoshenko:
@climba03003 @Eomm @RafaelGSS PTAL if you have some time.

climba03003 (approved):
LGTM, it retains backward compatibility

lib/constrainer.js
```diff
@@ -38,19 +45,30 @@ class Constrainer {
       throw new Error(`There already exists a custom constraint with the name ${strategy.name}.`)
     }
 
-    if (this.strategiesInUse.has(strategy.name)) {
+    if (this.isStrategyUsed(strategy.name)) {
       throw new Error(`There already exists a route with ${strategy.name} constraint.`)
     }
 
     strategy.isCustom = true
+    strategy.isAsync = strategy.deriveConstraint.length === 3
```
Eomm, line 53: As developer, I would expect something like `strategy.deriveConstraint.constructor.name === 'AsyncFunction'` here

As user, I would expect the support for `async` functions

I'm PRO with a followup PR, but it should be more clear for the reader

lib/constrainer.js
```diff
@@ -38,19 +45,30 @@ class Constrainer {
...
       this.noteUsage({ [strategy.name]: strategy })
     }
   }
 
-  deriveConstraints (req, ctx) {
+  deriveConstraints (req, ctx, done) {
+    const constraints = this.deriveSyncConstraints(req, ctx)
+
+    if (done === undefined) {
+      return constraints
```
Eomm, line 65: Sync contraints have priority over the async one?

As example:
https://github.com/Eomm/header-constraint-strategy#cases-explanation

> Multiple constraint matches

How do we want to deal the bullet point 7?

lib/constrainer.js
climba03003: `strategy.deriveConstraint.length === 3` means when the function argument is 3. Then, it is using `callback` which is `async`.

I don't expect `async function` here since `async-await` is slow for a router.

lib/constrainer.js
ivan-tymoshenko: I'm not sure that I understand you. You can have a mix of sync and async constraints.
https://github.com/delvedor/find-my-way/blob/2f233391cff8fdd6bdac57c737ea51ed015f8548/test/constraint.custom.async.test.js#L70

The `return` you ask for means that if you use sync deriveConstraint (sync lookup), you will get only sync constraints.

lib/constrainer.js
Eomm: I know it, but new devs mix callback and async every time, and they expect an `async function` out of the box.

I learnt to avoid such confusion 😂 

> is slow for a router.

image

climba03003: > I know it, but new devs mix callback and async every time

Then, improve the document is better choice. And we should explain why we decide not to have `async function` here.

lib/constrainer.js
Eomm: I mean this use case:

```
FindMyWay(
{ constraints: { 
   bar: asyncCustomResponseTypeStrategy,
   foo: oldCustomSyncStrategy
 } 
});
```

ivan-tymoshenko: Yes, this will work.

ivan-tymoshenko: It's correct behavior. With the `async: 'bar'` you didn't match the constraint, fallback to the defaultRouter, and you get this error because you pass `null` as a response.

Can you describe what is wrong and how do you want this to work?

Eomm: Sorry, I oversee the issue

I tested a bit more but as you said all constraints are running

Eomm (approved):

lib/constrainer.js
ivan-tymoshenko: @Eomm @climba03003 What did we decide? Are we good with the callback interface?

climba03003: > Are we good with the callback interface?

I believe yes, and we need a `caution` on the document about it doesn't support `async-await` and why we doesn't support it.

mcollina (approved):
lgtm

PR #312, opened 2022-12-20
fix: using pre-defined http methods instead http module

sokcuri:
I would like to use find-my-way in an environment other than Node.js environment (Hybrid, Browser, etc.)

There is no problem with other modules because browserify polyfills are implemented, but in the find-my-way library, METHODS is imported from the http module, and no alternative polyfills exist.

To work around that issue, declare METHODS as a constant within find-my-way instead of importing it from the http module.

### DIFF
--- a/lib/http-methods.js
+++ b/lib/http-methods.js
@@ -1,5 +1,6 @@
 'use strict'
 
+// defined by Node.js http module, a snapshot from Node.js 18.12.0
 const httpMethods = [
   'ACL', 'BIND', 'CHECKOUT', 'CONNECT', 'COPY', 'DELETE',
   'GET', 'HEAD', 'LINK', 'LOCK', 'M-SEARCH', 'MERGE',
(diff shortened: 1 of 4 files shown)


lib/http-methods.js
```diff
@@ -0,0 +1,10 @@
+const httpMethods = [
```
mcollina, line 1:
```suggestion
'use strict'

const httpMethods = [
```

mcollina (commented):
Good work. Note that there might need a bit more work on dependencies (assert and others) for browser usage.

sokcuri:
Other dependencies works fine with webpack (but a few configuration needs)

mcollina (approved):
lgtm

ivan-tymoshenko (approved):
lgtm

minidonut:
How about testing equality between `require('http').METHODS` and the hardcoded array `httpMethods`?

sokcuri:
I am aware that `require('http').METHODS` may vary depending on the Node.js version.
it will be difficult to test. Instead, I'll write down which snapshot I took as a comment.

PR #320, opened 2023-02-12
feat: add multiple parametric nodes support

### LINKED ISSUE #317: Allow users to have an additional regex node
find-my-way has a [caveat](https://github.com/delvedor/find-my-way#caveats): it can't have two different routes which differ only for their parameters. What we can allow users to have is a regular parametrical node and a regexp node (or multi-parametrical node, which is also a regexp node), in case there is only one regex node. In this case, we know that the regex node has higher priority than the regular one.

Example:
- `/foo/:param1`
- `/foo/:param2.png`

@mcollina @Eomm @Uzlopak @RafaelGSS @climba03003 WDYT?

ivan-tymoshenko:
Close #317

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,25 +283,34 @@ Having a route with multiple parameters may affect negatively the performance, s
 <a name="match-order"></a>
 ##### Match order
 
-The routing algorithm matches one chunk at a time (where the chunk is a string between two slashes),
+The routing algorithm matches one node at a time (where the node is a string between two slashes),
 this means that it cannot know if a route is static or dynamic until it finishes to match the URL.
 
-The chunks are matched in the following order:
+The nodes are matched in the following order:
 
 1. static
-1. parametric
-1. wildcards
-1. parametric(regex)
-1. multi parametric(regex)
+2. parametric node with static ending
+3. parametric(regex)/multi-parametric
+4. parametric
+5. wildcard
 
 So if you declare the following routes
 
-- `/:userId/foo/bar`
-- `/33/:a(^.*$)/:b`
+- `/foo/filename.png` - static route
+- `/foo/:filename.png` - route with param `filename` and static ending `.png`
+- `/foo/:filename.:ext` - route with two params `filename` and `ext`
+- `/foo/:filename` - route with one param `filename`
+- `/*` - wildcard route
 
-and the URL of the incoming request is /33/foo/bar,
-the second route will be matched because the first chunk (33) matches the static chunk.
-If the URL would have been /32/foo/bar, the first route would have been matched.
+You will have next matching rules:
+- the static node would have the highest priority. It will be matched only if incoming URL equals  `/foo/filename.png`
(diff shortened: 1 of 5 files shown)


mcollina (commented):
does this affects our benchmarks?

ivan-tymoshenko:
No, all changes are in the compile time.

ivan-tymoshenko:
This PR is ready to review. It allows you to have two identical routes with different param types: general parameter and regexp parameter.

Example:
- `/foo/:param`
- `/foo/:param(123)`

Code in the `index.js` is pretty tricky. I don't ask you to review it unless you have a lot of time to deal with it. Please take a look at two things:

1. Params [sorting function](https://github.com/delvedor/find-my-way/pull/320/files#diff-8370cfb1d23c8a7441ac8b34e2c5cca7ced2bc7bf84365b3dc6da8bd01d11f14R78). It describes how we would prioritese parametric nodes. (param node with long suffix -> param node with short suffix -> regexp node -> param node without suffix)

2. Please look at the test file. It more or less describes all params collisions and priorities. It's pretty simple and readable.

If you have any suggestions, you are more than welcome. 

cc @mcollina @climba03003 @Eomm

mcollina (changes requested):
Can you add some docs for this?

climba03003 (commented):
Can we defer the sorting to certain point later?
For example, once the first call of `lookup`.

It seems to do a lot of unnecessary sorting when there are many routes.

ivan-tymoshenko:
> Can we defer the sorting to certain point later? For example, once the first call of `lookup`.
> 
> It seems to do a lot of unnecessary sorting when there are many routes.

Technically we can, you are right. But you will have this sorting in a very small amount of cases. You a sorting parametric childs of _one_ parent node. IMHO, in 99% of all use cases, you have one parametric child, and in 1%, you have two children: general and regexp. All other cases are less than the statistical error.

On the other hand, this will complicate the already complicated code and introduce an additional error in measuring the hot path. I would agree with you 100% if it would make any difference, but you will not notice it.

climba03003:
> You a sorting parametric childs of one parent node. IMHO, in 99% of all use cases, you have one parametric child, and in 1%, you have two children. All other cases are less than the statistical error.

At least we can do `if(this.parametricChildren.length > 1)` to prevent single item array being sorted.

ivan-tymoshenko:
> At least we can do if(this.parametricChildren.length > 1) to prevent single item array being sorted.

I think that the sort function is smart enough to not call a sorting handler for an array with one item. But I can add this, not a problem at all.

delvedor (approved):
LGTM with some docs!

climba03003:
> > At least we can do if(this.parametricChildren.length > 1) to prevent single item array being sorted.
> 
> I think that the sort function is smart enough to not call a sorting handler for an array with one item. But I can add this, not a problem at all.

nvm, it is actually smart enough to skip the run.

```js
function sorter(a, b) {
  console.log('did run')
  return a - b
}

;['a'].sort(sorter)
;['a', 'b'].sort(sorter)
```

climba03003 (approved):

mcollina (approved):
lgtm

PR #324, opened 2023-04-08
Made the validate function of the ContraintStrategy in the typescript…

nickwinger:
… definition optional (as it is also optional in the js-code) added the validate function in the readme doc

### DIFF
--- a/README.md
+++ b/README.md
@@ -730,7 +730,10 @@ const customVersioning = {
   deriveConstraint: (req, ctx) => {
     return req.headers['accept']
   },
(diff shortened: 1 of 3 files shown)


ivan-tymoshenko (approved):
lgtm

mcollina:
Can you please add an assertion/example in https://github.com/delvedor/find-my-way/blob/main/test/types/router.test-d.ts? So we don't regress.

nickwinger:
> Can you please add an assertion/example in https://github.com/delvedor/find-my-way/blob/main/test/types/router.test-d.ts? So we don't regress.

ok, i made another commit

mcollina (approved):
lgtm

PR #328, opened 2023-05-27
Add package.json "files" entry

### LINKED ISSUE #327: Exclude test/ from package distribution
Can the `test/` dir be excluded from the npm module?

I found https://github.com/delvedor/find-my-way/issues/213 but it's a bit old and there doesn't seem to be a `.npmignore` any longer.

A better approach might be the [`"files"` key in `package.json`](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#files). This would be an "allowlist" and could help do away with other files like `benchmark/`

It would actually make a huge difference in package size 
image

tbeseda:
I realize using `"files"` in `package.json` essentially inverts how files are allowed into the published library, but it significantly reduces the size on disk for `find-my-way` (yes already small, but could be super small 😎 ).

```diff
npm notice === Tarball Details === 
npm notice name:          find-my-way                             
npm notice version:       7.6.2                                   
npm notice filename:      find-my-way-7.6.2.tgz                   
- npm notice package size:  79.4 kB
- npm notice unpacked size: 378.0 kB
- npm notice total files:   89
+ npm notice package size:  24.3 kB
+ npm notice unpacked size: 88.9 kB
+ npm notice total files:   14
```

The files that are packaged look correct to me, but here's a list

```
LICENSE                         
README.md                       
index.d.ts                      
index.js                        
lib/constrainer.js              
lib/handler-storage.js          
lib/http-methods.js             
lib/node.js                     
lib/pretty-print.js             
lib/strategies/accept-host.js   
lib/strategies/accept-version.js
lib/strategies/http-method.js   
lib/url-sanitizer.js            
package.json
```

fixes #327
Happy to hear feedback/make changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,11 @@
   "description": "Crazy fast http radix based router",
   "main": "index.js",
   "types": "index.d.ts",
+  "files": [
+    "index.js",
+    "index.d.ts",
+    "lib"
+  ],
   "scripts": {
     "bench": "node ./benchmark/bench.js",
     "bench:cmp": "node ./benchmark/compare-branches.js",


ivan-tymoshenko (approved):
Thanks for the PR. I like the idea of removing files from the package, but I would prefer to have an “exclude” behavior rather than “include”. Someone can add a file that would be not included, but tests would pass. @mcollina WDYT?

tbeseda:
I created #329 as an alternative that uses `.npmignore` 🍻

Eomm:
I have this saved quick answer about this topic:

> Thank you for your interest in this topic. Please see the discussion in https://github.com/fastify/skeleton/issues/42#issuecomment-1208090783 for my reasoning against such a change.

But this repo is @delvedor domain 🙌🏼

felixmosh:
Why not using `.npmignore` files to exclude `tests` / `benchmark` folders?

PR #331, opened 2023-08-09
chore: `npm run test` failing due to standard

calumrussell:
Cloned main today, tried `npm run test`, failed with a few errors that went away with `standard --fix` and one error that required change from `.map` to `.forEach`.

First contribution here but doesn't seem to have broken anything and  `npm run test` now running.


mcollina (approved):
lgtm

PR #334, opened 2023-08-30
docs: fix links

suruaku:
A few links were broken and a few were missing.
To be more consistent I'd make additional changes so it'd look like this:
- ab (link without prefix)
  - a-prefix-1
  - b-prefix-1
  - a-prefix-2
  - b-prefix-2


mcollina (approved):
lgtm

PR #336, opened 2023-09-24
feat: disable semicolon separator by default

levensta:
This PR comes from the discussion in the issue (https://github.com/fastify/fastify/issues/5050). I've added a new parameter that allows the `;` character to be used to separate path and query as before, but its default value is `false` to be fully compliant with the RFC

### DIFF
--- a/index.js
+++ b/index.js
@@ -88,6 +88,7 @@ function Router (opts) {
   this.maxParamLength = opts.maxParamLength || 100
   this.allowUnsafeRegex = opts.allowUnsafeRegex || false
   this.constrainer = new Constrainer(opts.constraints)
+  this.useSemicolonDelimiter = opts.useSemicolonDelimiter || false
 
   this.routes = []
   this.trees = {}
@@ -419,7 +420,7 @@ Router.prototype.find = function find (method, path, derivedConstraints) {
   let shouldDecodeParam
 
   try {
-    sanitizedUrl = safeDecodeURI(path)
+    sanitizedUrl = safeDecodeURI(path, this.useSemicolonDelimiter)
     path = sanitizedUrl.path
     querystring = sanitizedUrl.querystring
     shouldDecodeParam = sanitizedUrl.shouldDecodeParam
--- a/lib/url-sanitizer.js
+++ b/lib/url-sanitizer.js
@@ -62,7 +62,7 @@ function safeDecodeURI (path, useSemicolonDelimiter) {
     // Some systems do not follow RFC and separate the path and query
     // string with a `;` character (code 59), e.g. `/foo;jsessionid=123456`.
     // Thus, we need to split on `;` as well as `?` and `#` if the useSemicolonDelimiter option is enabled.
-    } else if (charCode === 63 || charCode === 35 || (useSemicolonDelimiter && charCode === 59)) {
+    } else if (charCode === 63 || charCode === 35 || (charCode === 59 && useSemicolonDelimiter)) {
       querystring = path.slice(i + 1)
       path = path.slice(0, i)
       break
--- a/lib/url-sanitizer.js
+++ b/lib/url-sanitizer.js
@@ -34,7 +34,7 @@ function decodeComponentChar (highCharCode, lowCharCode) {
   return null
 }
 
-function safeDecodeURI (path) {
+function safeDecodeURI (path, useSemicolonDelimiter) {
   let shouldDecode = false
   let shouldDecodeParam = false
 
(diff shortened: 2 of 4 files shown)


index.js
```diff
@@ -88,6 +88,7 @@ function Router (opts) {
   this.maxParamLength = opts.maxParamLength || 100
   this.allowUnsafeRegex = opts.allowUnsafeRegex || false
   this.constrainer = new Constrainer(opts.constraints)
+  this.useSemicolonDelimiter = opts.useSemicolonDelimiter || false
```
ivan-tymoshenko, line 91: Don't forget to set it `true` in the Fastify.

ivan-tymoshenko (approved):
This is really _weird_, but it looks like it's correct. Can you provide a bench results?
@mcollina This is a major for find-my-way.

lib/url-sanitizer.js
```diff
@@ -61,8 +61,8 @@ function safeDecodeURI (path) {
       }
     // Some systems do not follow RFC and separate the path and query
     // string with a `;` character (code 59), e.g. `/foo;jsessionid=123456`.
-    // Thus, we need to split on `;` as well as `?` and `#`.
-    } else if (charCode === 63 || charCode === 59 || charCode === 35) {
+    // Thus, we need to split on `;` as well as `?` and `#` if the useSemicolonDelimiter option is enabled.
+    } else if (charCode === 63 || charCode === 35 || (useSemicolonDelimiter && charCode === 59)) {
       querystring = path.slice(i + 1)
```
ivan-tymoshenko, line 66: If there is any perf difference (I doubt), you can move the `useSemicolonDelimiter`additional check under the `charCode` checking `if`. That would remove it from a hot path.

index.js
levensta: Ideally this option should be disabled in Fastify as well, but I don't know the best way to do it: leave the default `false` here and open issue/PR to update dependencies in Fastify, or make it `true` here and send PR to Fastify with the option disabled

ivan-tymoshenko: I would leave the `false` here, because it's a correct behavior and we can release a new major of fmw at any moment. In current version of Fastify it should be set to `true` by default, because otherwise it's a breaking change.

About changing it to `false` in the fastify in a future release. It's better to open an issue about this in fastify, but IMHO it never would be set to `true` by defauld in the Fastify, because it will break too many apps. It just doesn't worth it even if it's true according to the spec.

lib/url-sanitizer.js
levensta: Yes, given that the `;` symbol is rare, that's correct. I fixed

levensta:
> Can you provide a bench results?

I made a simple example in the sandbox
https://codesandbox.io/p/sandbox/find-my-way-semicolon-g29hw7

btw, it works the same way for the `#` character, but I also don't understand for which situation this is necessary. At the same time in Fastify the hash from request does not get into query object

mcollina (approved):
lgtm

levensta:
> @levensta did you run the benchmarks by any chance?

I run `bench:cmp` and got these results 🤔

<details>

<summary>screenshot</summary>

Снимок экрана 2023-09-26 в 17 57 18

</details>

mcollina (changes requested):
Can you document the new option?

mcollina:
I'm lost. can you make an example?

levensta:
> I'm lost. can you make an example?

In this example, you can see that both `;` and `#` work as delimiters
https://codesandbox.io/p/sandbox/find-my-way-semicolon-g29hw7?file=%2Findex.js%3A7%2C1

mcollina:
That does not look correct, what does the rfc say?

levensta:
Sorry for the long response time. The RFC explicitly says that the "#" character is used as the beginning of a separate segment called a fragment

> [3.  Syntax Components](https://www.rfc-editor.org/rfc/rfc3986#section-3)
> ```
> URI         = scheme ":" hier-part [ "?" query ] [ "#" fragment ]
> ```

> [3.3.  Path](https://www.rfc-editor.org/rfc/rfc3986#section-3.3)
> The path is terminated by the first question mark ("?") or number sign ("#") character, or by the end of the URI.

The problem with the current code is that it truncates the path to the delimiter character, but does not check which one is a query string and which one is a fragment and writes everything into one `querystring` variable. Judging by the history, this behaviour was originally https://github.com/delvedor/find-my-way/commit/a1cc69ae5909357c6e8421a85db526dcf13b489b
It's just that nothing was written to a separate `querystring` variable before.


A fix could look like this
image

Should this behaviour be hidden behind the same parameter as `useSemicolonDelimiter` (renamed in this case) or a other parameter? Or maybe we shouldn't hide this behaviour at all? It would be strange if any of the library's consumers were intentionally targeting this behaviour

Uzlopak:
It is correct that # should split. It is the "anker" in browser url, where you jump to the specific position.

The correct behavior is to ignore everything after a #.

So in conclusion: everything after the first occurence of `?` and the first occurence of  `#` is a querystring.

mcollina (approved):
lgtm

PR #337, opened 2023-09-24
feat: add findRoute and hasRoute methods

### LINKED ISSUE #332: Add `has` method
See https://github.com/fastify/fastify/pull/4986#issuecomment-1687015746

The `has` method should parse the given path in the same way as [_on](https://github.com/delvedor/find-my-way/blob/ec25619ac06c16f9aabcab54995ce96f91390a62/index.js#L142) method does.

1. It shouldn't modify the the existing tree structure or has any other side effects.
2. Instead of [throwing an error](https://github.com/delvedor/find-my-way/blob/ec25619ac06c16f9aabcab54995ce96f91390a62/index.js#L286) if path exists, the `has` method should return a boolean value here.
3. It shouldn't be a copy/paste. `_on` and `has` methods should reuse the same code or even each other.

ivan-tymoshenko:
Close #332
Issue in fastify: https://github.com/fastify/fastify/issues/4866

### DIFF
--- a/README.md
+++ b/README.md
@@ -389,6 +389,51 @@ router.off(['POST', 'GET'], '/example', { host: 'fastify.io' })
 router.off(['POST', 'GET'], '/example', {})
 ```
 
+#### findRoute (method, path, [constraints])
+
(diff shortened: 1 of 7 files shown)


mcollina:
What's missing?
Could you skip Node.js v14 on Windows?

mcollina:
please drop windows and node v14, it's broken on Gha

mcollina (approved):
lgtm

PR #338, opened 2023-10-05
ci: update nodejs versions

ivan-tymoshenko:


.github/workflows/node.js.yml
```diff
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x, 16.x, 18.x]
+        node-version: [16.x, 18.x, 20.x]
```
mcollina, line 15: We need to keep supporting Node v14 for supporting Fastify v4. Just exclude Windows v14 which is broken on GHA.

mcollina (changes requested):

.github/workflows/node.js.yml
ivan-tymoshenko: we still suuport v14? 🤦🏻

mcollina: we need to plan fastify v5

mcollina (approved):
lgtm

Issue #343, opened 2024-01-15
Host constraint breaks with 32 handlers


valeneiko:
If I add 32 handlers with host constraints, then none of the handlers will match and no errors are thrown. 

```js
const router = require('find-my-way')();

const N = 32;

for (let i = 0; i < N; i++) {
  const host = `h${i.toString().padStart(2, '0')}`;
  router.on('GET', '/', { constraints: { host } }, () => { });
}

const result = router.find('GET', '/', { host: 'h01' });
console.log(result); // --> null
```

If I change `N` to `33` then error is thrown from the code below:
https://github.com/delvedor/find-my-way/blob/ca54d8df63efe051c363f29a99aa81c7d9e76fb5/lib/handler-storage.js#L50-L52

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

PR #344, opened 2024-01-16
Update contraint check to throw with 32 handlers

valeneiko:
Fixes:
- #343

### DIFF
--- a/lib/handler-storage.js
+++ b/lib/handler-storage.js
@@ -47,8 +47,8 @@ class HandlerStorage {
     }
 
     const isMergedTree = constraintsNames.includes(httpMethodStrategy.name)
-    if (!isMergedTree && this.handlers.length >= 32) {
-      throw new Error('find-my-way supports a maximum of 32 route handlers per node when there are constraints, limit reached')
+    if (!isMergedTree && this.handlers.length >= 31) {
+      throw new Error('find-my-way supports a maximum of 31 route handlers per node when there are constraints, limit reached')
     }
 
     this.handlers.push(handlerObject)
--- a/test/constraint.host.test.js
+++ b/test/constraint.host.test.js
@@ -74,3 +74,32 @@ test('A route supports multiple host constraints (lookup)', t => {
     headers: { host: 'bar.fancy.ca' }
   })
 })
+
+test('A route supports up to 31 host constraints', (t) => {
+  t.plan(1)
+
+  const findMyWay = FindMyWay()
+
+  for (let i = 0; i < 31; i++) {
+    const host = `h${i.toString().padStart(2, '0')}`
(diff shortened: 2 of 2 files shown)


lib/handler-storage.js
```diff
@@ -47,8 +47,8 @@ class HandlerStorage {
     }
 
     const isMergedTree = constraintsNames.includes(httpMethodStrategy.name)
-    if (!isMergedTree && this.handlers.length >= 32) {
-      throw new Error('find-my-way supports a maximum of 32 route handlers per node when there are constraints, limit reached')
+    if (!isMergedTree && this.handlers.length >= 31) {
+      throw new Error('find-my-way supports a maximum of 31 route handlers per node when there are constraints, limit reached')
```
mcollina, line 51:
```suggestion
    if (!isMergedTree && this.handlers.length >= 33) {
      throw new Error('find-my-way supports a maximum of 32 route handlers per node when there are constraints, limit reached')
```

test/constraint.host.test.js
```diff
@@ -74,3 +74,32 @@ test('A route supports multiple host constraints (lookup)', t => {
     headers: { host: 'bar.fancy.ca' }
   })
 })
+
+test('A route supports up to 31 host constraints', (t) => {
+  t.plan(1)
+
+  const findMyWay = FindMyWay()
+
+  for (let i = 0; i < 31; i++) {
```
mcollina, line 83:
```suggestion
  for (let i = 0; i < 32; i++) {
```

test/constraint.host.test.js
```diff
@@ -74,3 +74,32 @@ test('A route supports multiple host constraints (lookup)', t => {
...
+
+  t.equal(findMyWay.find('GET', '/', { host: 'h01' }).handler, alpha)
+})
+
+test('A route throws when constraint limit exceeded', (t) => {
+  t.plan(1)
+
+  const findMyWay = FindMyWay()
+
+  for (let i = 0; i < 31; i++) {
```
mcollina, line 96:
```suggestion
  for (let i = 0; i < 33; i++) {
```

lib/handler-storage.js
valeneiko: Isn't this the opposite of what we want? 

The logic on main throws when we add 33 handlers, but routing breaks with 32.

So what I was trying to achieve with changing to 31 is to throw when we know routing breaks. The reason we need to use 31 here, is because the check is done before we have added current handler to the list.

mcollina:
@ivan-tymoshenko can you take a look?

ivan-tymoshenko (approved):
Indeed it doesn't work for 32 constrainst values, for multiple reasons. One of them is that indexing constraints values from `1`.

PR #345, opened 2024-01-31
Fix header in README

selimb:
The `lookup` header was mistakenly wrapped with ` ```js `.

Also checked all other links in the table of contents.


mcollina (approved):
lgtm

PR #347, opened 2024-02-16
feat: add node: prefix for assert

### LINKED ISSUE #346: Incompatibility with Cloudflare due to using assert instead of node:assert
Cloudflare has support for some Node.js packages ([link](https://developers.cloudflare.com/workers/runtime-apis/nodejs/)). There is support for `assert` package, but as I understand it only with the use of `node:` prefix (like `node:assert`). I am not an expert in Node.js , but maybe there is a possibility to add this prefix? I would be happy to do it myself if this option is suitable for CloudFlare support of your codebase.

SavaCool122:
Resolves #346

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,7 +25,7 @@
     '~': 126 - ~
 */
 
-const assert = require('assert')
+const assert = require('node:assert')
 const querystring = require('fast-querystring')
 const isRegexSafe = require('safe-regex2')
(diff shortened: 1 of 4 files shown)


ivan-tymoshenko (approved):

mcollina (approved):
lgtm

---

I'm somewhat suprised this runs everywhere correctly, it might have shipped sooner then! Time flies

PR #383, opened 2024-12-05
perf: use map for sting-based stores

gurgunday:
When an object is used as a map and has string-based keys, Map's performance will be superior

For this repo, not sure how to test it but here's some related discussion and benchmarks in https://github.com/kibertoad/toad-cache/issues/40

And a V8 engineer explaining why: https://stackoverflow.com/questions/62350146/why-map-manipulation-is-much-slower-than-object-in-javascript-v8-for-integ/62351925#62351925

### DIFF
--- a/lib/strategies/accept-host.js
+++ b/lib/strategies/accept-host.js
@@ -2,11 +2,11 @@
 const assert = require('node:assert')
 
 function HostStorage () {
-  const hosts = {}
+  const hosts = new Map()
   const regexHosts = []
   return {
     get: (host) => {
-      const exact = hosts[host]
+      const exact = hosts.get(host)
       if (exact) {
         return exact
       }
@@ -20,7 +20,7 @@ function HostStorage () {
       if (host instanceof RegExp) {
         regexHosts.push({ host, value })
       } else {
-        hosts[host] = value
+        hosts.set(host, value)
       }
     }
   }
--- a/lib/strategies/accept-version.js
+++ b/lib/strategies/accept-version.js
@@ -7,8 +7,7 @@ function SemVerStore () {
     return new SemVerStore()
   }
 
-  this.store = {}
-
+  this.store = new Map()
   this.maxMajor = 0
   this.maxMinors = {}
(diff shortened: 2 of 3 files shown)


mcollina:
Did you run the benchmarks? Where there any improvements?

gurgunday:
I will post a benchmark if I can

gurgunday:
Yeah no significant improvement here

gurgunday:
@mcollina so I ran some tests again

For keys that aren't integers, Map is still consistently faster, it also doesn't force V8 to reconsider the hidden class a bunch of times

We had the same results in toad-cache, so it's perfectly safe to land

We don't have a benchmark that tests these because they are only used in `Constrainer`, but here is a generic benchmark:

```sh
Map time for 1000 iterations: 0.90 ms
Object time for 1000 iterations: 1.60 ms
```

```js
function bench(iterations) {
    const map = new Map();
    const obj = {};

    for (let i = 1; i <= 10; i++) {
        map.set(`${i}.x.x`, `host${i}`);
        obj[`${i}.x.x`] = `host${i}`;
    }

    function benchmarkMap() {
        let sum = 0;
        for (let i = 0; i < iterations; i++) {
            for (let j = 1; j <= 10; j++) {
                sum += map.get(`${j}.x.x`).length;
            }
        }
        return sum;
    }

    function benchmarkObject() {
        let sum = 0;
        for (let i = 0; i < iterations; i++) {
            for (let j = 1; j <= 10; j++) {
                sum += obj[`${j}.x.x`].length;
            }
        }
        return sum;
    }

    const startMap = performance.now();
    benchmarkMap();
    const endMap = performance.now();

    const startObject = performance.now();
    benchmarkObject();
    const endObject = performance.now();

    console.log(`Map time for ${iterations} iterations:`, (endMap - startMap).toFixed(2), "ms");
    console.log(`Object time for ${iterations} iterations:`, (endObject - startObject).toFixed(2), "ms");
}

bench(1000);
```

mcollina (approved):
lgtm

PR #389, opened 2025-02-28
fix: missing the param done of lookup method

zhennann:
missing the param done of lookup method


mcollina (changes requested):
Can you add a test? We use tsd for testing types

mcollina (approved):
lgtm

PR #401, opened 2025-06-16
ci: use tags for immutable github actions

Fdawgs:
Most first party actions now use https://github.com/actions/publish-immutable-action, which negates the need for commit hashes.


mcollina (approved):
lgtm

PR #423, opened 2026-04-19
fix: default pathOptional to '/' when deregistering root optional params

deepview-autofix:
The `off` method replaced the optional parameter in the path but did not fall back to '/' when the resulting string was empty, unlike `on`. For paths like `/:param?`, `off` threw "The path could not be empty" instead of removing the route registered at '/'.


ChALkeR (commented):
the test was throwing on `findMyWay.off('GET', '/:optional?')` before the fix

mcollina (approved):
lgtm

mcollina:
@deepview-autofix @ChALkeR can you rebase this?

mcollina:
Superseded by #428, which already merged the equivalent fix.

PR #424, opened 2026-04-19
fix: align isParamSafe state in findRoute with _on

deepview-autofix:
In findRoute, isParamSafe was set to false after processing a regex parameter, while _on sets it to true in the same position. The inconsistency caused the two methods to generate different regexes for routes that mix regex and non-regex params in one segment (e.g. '/:name(\\d+)-:other'), so getParametricChild could not match the node registered by _on and findRoute/hasRoute returned null.

### DIFF
--- a/index.js
+++ b/index.js
@@ -378,7 +378,7 @@ Router.prototype.findRoute = function findNode (method, path, constraints = {}) 
             regexps.push(trimRegExpStartAndEnd(regexString))
 
             j = endOfRegexIndex + 1
-            isParamSafe = false
+            isParamSafe = true
           } else {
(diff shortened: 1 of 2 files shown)


ChALkeR (commented):
In main:
https://github.com/delvedor/find-my-way/blob/eb8ec6fb65ace34435ba766f786371f09b020b75/index.js#L233-L238

But:
https://github.com/delvedor/find-my-way/blob/eb8ec6fb65ace34435ba766f786371f09b020b75/index.js#L380-L385

Since https://github.com/delvedor/find-my-way/commit/5e9e0eb5d8d438e06a185d5e536a896572dd0440

This change should be safe as it only affects `findRoute` which does not run the produced regexes, just checks if the router has them, and in `.on` it was already `isParamSafe = true`.

mcollina (approved):
lgtm

PR #425, opened 2026-04-19
fix: prevent done from being called multiple times on async constraint errors

deepview-autofix:
When multiple async constraint strategies are in use and more than one calls back with an error, `deriveAsyncConstraints` would invoke the `done` callback once per error. Track an `errored` flag so the first error short-circuits subsequent callbacks and `done` is only invoked once.


mcollina (approved):
lgtm

Issue #330, opened 2023-06-08
Get Test Coverage to 100%


ivan-tymoshenko:

ivan-tymoshenko:
@Ekott2006 Hi, I see you are interested in contributing to the project. I really appreciate it. Unfortunately, most of the issues that you can find in the backlog are stuck because they don't have a good answer or the solution doesn't worth it. You can start with this issue. We have a few uncovered lines in the project. Some of them are pretty simple, and some of them require investigation. It's a good issue problem to get familiar with the project. Tell me if you are interested, and I will assign it to you.

Ekott2006:
Let me be honest, I have never used find-my-way before but I am interested to contribute to open source nodejs project. So what exactly am I testing?

ivan-tymoshenko:
you can run `npm test` command. at the end, you will see a test coverage report. There are some lines that are not covered with tests (they marks with red color). You can create a test for them if it's possible or find the reason why they can't be covered (it's possible that there is some redundant code).

Ekott2006:
No problem, I will come back when I face isssues

jean-michelet:
I can give it a try if you want @ivan-tymoshenko

ivan-tymoshenko:
> I can give it a try if you want @ivan-tymoshenko

Go ahead. This is still an actual issue.

PR #426, opened 2026-04-19
fix: preserve unrecognized %XX sequences in safeDecodeURIComponent

deepview-autofix:
`decodeComponentChar` returns `null` for percent-encoded sequences that are not reserved characters. Previously, `safeDecodeURIComponent` concatenated this `null` directly, producing the literal string `"null"` in the decoded output (e.g. a double-encoded `%2520` became `"null"` after `safeDecodeURI`/`safeDecodeURIComponent`). Fall back to the original `%XX` substring when decoding is not applicable.

### DIFF
--- a/lib/url-sanitizer.js
+++ b/lib/url-sanitizer.js
@@ -94,7 +94,7 @@ function safeDecodeURIComponent (uriComponent) {
       const lowCharCode = uriComponent.charCodeAt(i + 2)
 
       const decodedChar = decodeComponentChar(highCharCode, lowCharCode)
-      decoded += uriComponent.slice(lastIndex, i) + decodedChar
+      decoded += uriComponent.slice(lastIndex, i) + (decodedChar !== null ? decodedChar : uriComponent.slice(i, i + 3))
 
       lastIndex = i + 3
     }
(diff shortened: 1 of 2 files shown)


deepview-autofix (commented):
Draft, see comment below

ChALkeR (commented):
Of this one, I'm unsure

One one hand, `safeDecodeURI` is documented to preserve chars:
https://github.com/delvedor/find-my-way/blob/eb8ec6fb65ace34435ba766f786371f09b020b75/lib/url-sanitizer.js#L38

On the other hand, there is a test that expects `"null"` string in output:

https://github.com/delvedor/find-my-way/blob/eb8ec6fb65ace34435ba766f786371f09b020b75/test/issue-330.test.js#L164

That test was introduced in #330 though and could have been not by design but by documenting existing behavior / increasing coverage.

But anyway, the behavior should be either documented if it is by design or fixed if it's a bug

While this PR assumes the latter, it could be the former - not sure

mcollina:
The test was only added to increase coverage, I think should be fixed

ChALkeR:
Will recheck the logic behind decoding one more time as a safety measure

PR #427, opened 2026-04-29
fix: backtrack on regex param mismatch

mcollina:
## Summary

Fix regex param routes so they do not match requests with an empty trailing segment.

For example, `/users/:userId(^\\d+)` should not match `/users/`.

## Changes

- add a regression test for the empty trailing segment case
- backtrack when a regex param candidate fails to match instead of leaving lookup on the failed leaf node
- preserve existing fallback behavior for sibling parametric and wildcard routes

## Validation

- `npm test`

## Related

- fastify/fastify#6700

### DIFF
--- a/benchmark/compare-branches.js
+++ b/benchmark/compare-branches.js
@@ -2,7 +2,7 @@
 
 const { spawn } = require('child_process')
 
-const chalk = require('chalk')
+const { default: chalk } = require('chalk')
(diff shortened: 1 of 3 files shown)


ivan-tymoshenko (approved):
Tbh it's hard to say if the logic is correct. I didn't touch this code for a while.

The test looks correct though

PR #428, opened 2026-04-29
fix: apply root path fallback for optional params in off()

aquie00t:
## Problem

There is an inconsistency between how optional parameters are expanded in `on()` and `off()`.

When registering a route like `/:id?`, `on()` expands it into both `/:id` and `/`. However, `off()` does not mirror this behavior consistently.

In some edge cases, this may lead to `off()` being invoked with an empty path (`''`), which violates the path validation (`path.length > 0`).

---

## Root Cause

This behavior was already fixed for `on()` in commit cce5437 (#367), but `off()` was not updated with the same fallback:

```js
// on() — correct
const pathOptional = path.replace(OPTIONAL_PARAM_REGEXP, '$2') || '/'
```

```js
// off() — missing fallback
const pathOptional = path.replace(OPTIONAL_PARAM_REGEXP, '$2')
```

---

## Fix

Apply the same fallback used in `on()`:

```js
const pathOptional = path.replace(OPTIONAL_PARAM_REGEXP, '$2') || '/'
```

This ensures that optional parameters resolving to root (`''`) are normalized to `/` in both cases.

---

## Reproduction

```js
const findMyWay = require('find-my-way')()

findMyWay.on('GET', '/:id?', () => {})

findMyWay.off('GET', '/:id?')
```

---

## Result

`off()` becomes consistent with `on()` and correctly handles optional parameters that resolve to root.

### DIFF
--- a/index.js
+++ b/index.js
@@ -487,7 +487,7 @@ Router.prototype.off = function off (method, path, constraints) {
     assert(path.length === optionalParamMatch.index + optionalParamMatch[0].length, 'Optional Parameter needs to be the last parameter of the path')
 
     const pathFull = path.replace(OPTIONAL_PARAM_REGEXP, '$1$2')
-    const pathOptional = path.replace(OPTIONAL_PARAM_REGEXP, '$2')
+    const pathOptional = path.replace(OPTIONAL_PARAM_REGEXP, '$2') || '/'
 
     this.off(method, pathFull, constraints)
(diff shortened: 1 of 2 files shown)


aquie00t:
Hey @mcollina — this is directly related to your fix in #367. The same` || '/' `fallback was applied to `on()` but `off()` was missed. Would you mind taking a look?

mcollina (approved):
lgtm

PR #431, opened 2026-05-02
fix: return limitExceeded objext when maxParamLength is exceeded

sairam676:
## Problem
When a URL parameter exceeds `maxParamLength`, the router returns `null`.
Fastify interprets this as route not found (404) instead of URI Too Long (414).
Related: fastify/fastify#6697

## Fix
Changed `return null` to `return { limitExceeded: true }` in both 
maxParamLength checks in index.js (lines 675 and 681).
This allows Fastify to distinguish between "not found" and "param too long"
and respond with the correct 414 status code.

## Changes
- index.js: return { limitExceeded: true } instead of null
- test/issue-330.test.js: updated assertion to match new behaviour
- test/max-param-length.test.js: updated assertion to match new behaviour
- test/max-param-length-414.test.js: new test for this specific fix

### DIFF
--- a/index.js
+++ b/index.js
@@ -672,13 +672,13 @@ Router.prototype.find = function find (method, path, derivedConstraints) {
       for (let i = 1; i < matchedParameters.length; i++) {
         const matchedParam = matchedParameters[i]
         if (matchedParam.length > maxParamLength) {
-          return null
+          return { limitExceeded: true }
         }
         params.push(matchedParam)
       }
     } else {
       if (param.length > maxParamLength) {
-        return null
+        return { limitExceeded: true }
       }
       params.push(param)
     }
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -672,13 +672,13 @@ Router.prototype.find = function find (method, path, derivedConstraints) {
       for (let i = 1; i < matchedParameters.length; i++) {
         const matchedParam = matchedParameters[i]
         if (matchedParam.length > maxParamLength) {
-          return null
+          return { limitExceeded: true }
```
Eomm: This seems a breaking change, right now the pattern is:
- object: match found
- null: route not found

I think we should not return a simple object here, but an `Error()` or something to inspect: https://github.com/fastify/fastify/blob/11dc5f21c5d13f596bea2fd653db384bbb0764c2/lib/route.js#L184

sairam676: Thanks for the feedback! Updated to return a error instance instead of a plain object to avoid breaking the existing null/object contract. Tests updated accordingly.

PR #433, opened 2026-05-08
fix: ensure parametric routes require at least one character (#422)

### LINKED ISSUE #422: Parametric route /:id incorrectly matches / when no static route is registered
I have this setup:

```js
import http from "http";
import FindMyWay from "find-my-way";

const router = FindMyWay({
  defaultRoute: (req, res) => {
    res.statusCode = 404;
    res.end(JSON.stringify({ message: "Not Found" }));
  },
});

router.on("GET", "/:id", (req, res, params) => {
  res.end(JSON.stringify(params));
});

  router.lookup(req, res);
});

server.listen(3000);
```

When I hit `/`, I expect `defaultRoute` to handle it since there's no `/` route registered. Instead my `/:id` handler fires with `params.id = undefined`.

```bash
curl http://localhost:3000
# Expected: 404 from defaultRoute
# Actual:   {} — /:id handler called with id = undefined
```

The only workaround I found is manually registering a `/` route, which feels wrong when `defaultRoute` already exists for this purpose.

- `find-my-way` version: ^9.5.0
- Node.js version: v24.14.1

oldregime:
This PR fixes issue #422 where parametric routes (e.g. `/:id`) incorrectly matched the root path (`/`) with an empty parameter.

### Changes:
- Modified `getNextNode` in `lib/node.js` to ensure parametric children are only considered if there are characters remaining in the path.
- Updated tests to reflect this change. Parametric routes now require at least one character to match.
- Users who need to match empty values should use optional parameters (e.g. `/:id?`) or `ignoreTrailingSlash: true` if the empty value is represented by a trailing slash.

This aligns the behavior with modern HTTP routing standards and prevents unexpected matches to the root path when only parametric routes are registered.

Fixes #422

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -156,7 +156,7 @@ class StaticNode extends ParentNode {
     let parametricBrotherNodeIndex = 0
 
     if (node === null) {
-      if (this.parametricChildren.length === 0) {
+      if (this.parametricChildren.length === 0 || pathIndex === path.length) {
         return this.wildcardChild
       }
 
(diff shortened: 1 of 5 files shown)


mcollina:
I think this would be a semver-major change, will do it for Fastify 6

PR #436, opened 2026-05-19
perf(host): cache regex lookup results in HostStorage

aquie00t:
## Problem

Every `get()` call scans all regex patterns from scratch, even for
hosts seen before. In applications with regex-based host constraints,
this becomes a hot path with no reuse.

## Solution

Add a `Map`-based cache inside `HostStorage`. Once a host is matched
(or confirmed unmatched) against the regex list, the result is stored
and returned directly on subsequent calls.

An early-exit guard (`regexHosts.length === 0`) ensures exact-match
paths are completely unaffected.

The cache is invalidated when a new regex route is added via `set()`.

## Benchmark

500 samples, Node.js v26.1.0

| Scenario | Original | Patched | Delta |
|---|---|---|---|
| Exact match only | 38.0M ops/sec | 38.4M ops/sec | ~0% |
| Single regex, repeated host | 13.9M ops/sec | 43.4M ops/sec | +210% |
| Many regexes, finite host pool | 8.1M ops/sec | 34.0M ops/sec | +318% |

### DIFF
--- a/lib/strategies/accept-host.js
+++ b/lib/strategies/accept-host.js
@@ -4,21 +4,28 @@ const assert = require('node:assert')
 function HostStorage () {
   const hosts = new Map()
   const regexHosts = []
+  const regexCache = new Map()
   return {
     get: (host) => {
       const exact = hosts.get(host)
       if (exact) {
         return exact
       }
+      if (regexHosts.length === 0) return undefined
+      if (regexCache.has(host)) return regexCache.get(host)
       for (const regex of regexHosts) {
         if (regex.host.test(host)) {
+          regexCache.set(host, regex.value)
           return regex.value
         }
       }
+      regexCache.set(host, undefined)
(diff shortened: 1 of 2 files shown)


aquie00t:
@mcollina ping

lib/strategies/accept-host.js
```diff
@@ -4,21 +4,27 @@ const assert = require('node:assert')
 function HostStorage () {
   const hosts = new Map()
   const regexHosts = []
+  const regexCache = new Map()
```
mcollina, line 7: unforunately regexp are stateful in JavaScript :(

aquie00t: You're right, fixed by stripping `g`/`y` flags on insert. Added a test too.

mcollina (approved):
lgtm

PR #441, opened 2026-07-09
fix: avoid crash on undefined regex capture during route lookup

bianyifan:
This PR fixes a runtime crash in regex param route matching when a capture group is undefined.

In some routes using nested or repeated regex groups, a request can produce an undefined capture value while regex matching still succeeds. The lookup path then attempts to read .length from that undefined value, causing a TypeError and crashing the process.

What this PR changes:

1. Normalize regex capture values to empty strings before length checks.
2. Push normalized values into params, so params remain stable and lookup does not throw.

Behavior note:
This change preserves empty-value matching behavior. Empty captures are treated as empty strings rather than non-matches.

### DIFF
--- a/index.js
+++ b/index.js
@@ -690,7 +690,7 @@ Router.prototype.find = function find (method, path, derivedConstraints) {
 
         let regexMaxParamLengthExceeded = false
         for (let i = 1; i < matchedParameters.length; i++) {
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #442, opened 2026-07-24
fix: keep static parts between parameters in the route pattern

### LINKED ISSUE #369: Bug - parameter separator logic when path end of variable params
There are two urls
/h5/:cityName-:poiName/hotels-c:cityId-r:poiId 
/h5/:cityName-:poiName/hotels-c:cityId-a:poiId 

The different part are the first letter of "r:poiId" and "a:poiId"
Under normal circumstances, the two URLs above should not be recognized as two different URLs.

However find-my-way identify two urls have same pattern /h5/()/hotels-c() and then throw `Method 'get' already declared for route '/h5/()/hotels-c().' with constraints '{}'

After studied and debugged the source code and thought the problem should be in [L246](https://github.com/delvedor/find-my-way/blob/main/index.js#L246)
In line246 if the route is end with variable, find-my-way omissions the static part between two variable.

rajanpanth:
Fixes #369.

## Reproduction

The report in #369 never got a runnable reproduction, so here is one against `main` (v9.7.0):

```js
const FindMyWay = require('find-my-way')
const router = FindMyWay()

router.on('GET', '/h5/:cityName-:poiName/hotels-c:cityId-r:poiId', () => 'r')
router.on('GET', '/h5/:cityName-:poiName/hotels-c:cityId-a:poiId', () => 'a')
// Error: Method 'GET' already declared for route '/h5/:()/hotels-c:()' with constraints '{}'
```

The two routes differ in the literal `r` and `a` before the last parameter, so they are distinct routes, but the second registration is rejected as a duplicate.

The same root cause makes these collide too:

```js
router.on('GET', '/foo/:a-:b', () => 'dash')
router.on('GET', '/foo/:a.:b', () => 'dot')   // same error
```

## Cause

When a parametric node is closed, its canonical form is built from the most recent static part only:

```js
const nodePattern = isRegexNode ? '()' + staticPart : staticPart
```

For `:cityId-r:poiId` the loop sees two parameters. It records `-r` as the static part of the first, then the second parameter ends the node with an empty static part, so `staticPart` is `''` and the whole node canonicalises to `()`. Every static part between parameters is dropped whenever the node ends with a parameter — which is the diagnosis given in the issue.

The compiled regex is unaffected because it is accumulated across the loop into `regexps`, which is why matching and parameter extraction were always correct. Only pattern identity was wrong, and pattern identity is what the duplicate check at index.js:299 and `findRoute` at index.js:446 compare.

## Fix

Accumulate the canonical form one parameter at a time, so interior static parts survive:

```js
nodePatternParts += '()' + staticPart
...
const nodePattern = isRegexNode ? nodePatternParts : staticPart
```

`/h5/:cityName-:poiName/hotels-c:cityId-r:poiId` now canonicalises to `/h5/:()-()/hotels-c:()-r()` instead of `/h5/:()/hotels-c:()`.

Applied to both `_on` and `findRoute`, which contain the same parsing loop — if only one were changed, registration and lookup would disagree about a route's identity.

`pattern` is compared only against other generated patterns and is not part of the documented API, so the changed canonical form is internal.

## Tests

New `test/issue-369.test.js` covers the reported case, the `-` vs `.` variant, `findRoute` symmetry (including a negative lookup), that parameters still parse correctly, and that genuine duplicates ending in a parameter are still rejected.

Against unmodified `main`, 3 of the 5 fail; the other 2 pass on both and guard against regressing existing behaviour.

- `npx borp` — 519 passing (514 before this branch), 0 failing
- `npx standard` — clean
- `npm run test:typescript` — clean

### DIFF
--- a/index.js
+++ b/index.js
@@ -206,6 +206,9 @@ Router.prototype._on = function _on (method, path, opts, handler, store) {
       let isParamSafe = true
       let backtrack = ''
       const regexps = []
+      // Canonical form of this node, accumulated one parameter at a time so
+      // static parts between parameters survive into the pattern.
+      let nodePatternParts = ''
 
       let lastParamStartIndex = i + 1
       for (let j = lastParamStartIndex; ; j++) {
@@ -257,9 +260,10 @@ Router.prototype._on = function _on (method, path, opts, handler, store) {
           }
 
           lastParamStartIndex = j + 1
+          nodePatternParts += '()' + staticPart
 
           if (isEndOfNode || pattern.charCodeAt(j) === 47 || j === pattern.length) {
-            const nodePattern = isRegexNode ? '()' + staticPart : staticPart
+            const nodePattern = isRegexNode ? nodePatternParts : staticPart
             const nodePath = pattern.slice(i, j)
 
             pattern = pattern.slice(0, i + 1) + nodePattern + pattern.slice(j)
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #447, opened 2026-08-09
perf: faster route lookup (+18.7% geomean, every benchmark scenario faster)

mcollina:
## Summary

Three optimizations to the lookup hot path, no API changes:

**URL sanitizing (`lib/url-sanitizer.js`)** — `safeDecodeURI` scanned every URL char-by-char in JS on every lookup. Paths of 12+ chars now locate `%` and the query delimiters (`?`, `#`, and `;` when `useSemicolonDelimiter` is enabled) with native `String#indexOf` and return early when the path portion contains no percent-encoding, which is the common case. Shorter paths keep a char scan, split into a small function V8 reliably inlines into `find()` plus a percent-decoding slow path that only runs when a `%` is present. To keep the percent path itself inlineable (V8 counts a callee's already-inlined bytecode against its inlining limit), `decodeComponentChar`'s if-chain became a 768-byte lookup table with an integer-returning variant used on the scan path, and the rare `%25` re-encoding moved to its own function.

**Static child lookup (`lib/node.js`)** — static children were stored in an object keyed by the child's first character, so every tree level paid a dictionary lookup by single-char string (`KeyedLoadIC` was ~10% of profile ticks). They are now stored in parallel arrays keyed by charCode with a linear scan over integers.

**GET tree fast path (`index.js`)** — `trees` holds one key per HTTP method, and 35 keys is beyond V8's fast-properties limit, so it is a dictionary-mode object and every `trees[method]` access goes through a generic IC call (~11% of cycles on short static lookups). A dedicated `_treeGET` instance field turns the tree load for GET requests — the vast majority of real traffic — into a pointer compare plus a fixed-offset field load. Other methods use `trees` as before.

## Benchmarks

`benchmark/bench.js` on `main` vs this branch (Node v24.18.0, Linux x64). To control for CPU frequency drift between runs, both suites were run twice interleaved (main, PR, main, PR) taking the best per scenario. **Every scenario is faster**, and the short routes that dominate real traffic gain 3-13%:

| Scenario | main | this PR | Δ |
|---|---|---|---|
| lookup root "/" route | 16.32M | 17.66M | **+8.2%** |
| lookup short static route | 9.65M | 10.27M | **+6.4%** |
| lookup long static route | 3.32M | 4.58M | **+38.1%** |
| lookup long static route (common prefix) | 2.26M | 3.73M | **+64.9%** |
| lookup short parametric route | 7.07M | 7.26M | +2.7% |
| lookup long parametric route | 4.27M | 6.35M | **+48.8%** |
| lookup short parametric route (encoded unoptimized) | 4.34M | 4.37M | +0.5% |
| lookup short parametric route (encoded optimized) | 3.05M | 3.07M | +0.6% |
| lookup parametric route with two short params | 3.48M | 4.81M | **+38.3%** |
| lookup multi-parametric route with two short params | 3.09M | 3.19M | +2.9% |
| lookup multi-parametric route with two short regex params | 3.39M | 3.50M | +3.0% |
| lookup long static + parametric route | 1.93M | 2.82M | **+45.9%** |
| lookup short wildcard route | 7.89M | 8.45M | **+7.1%** |
| lookup long wildcard route | 4.54M | 7.37M | **+62.3%** |
| lookup root route on constrained router | 14.43M | 16.31M | **+13.0%** |
| lookup short static unconstraint route | 8.77M | 9.29M | **+5.9%** |
| lookup short static versioned route | 5.71M | 6.30M | **+10.4%** |
| lookup short static constrained (version & host) route | 5.80M | 6.35M | **+9.5%** |

**Geomean: +18.7%.** As a noise-free cross-check, `perf stat -e instructions` on fixed 10M-lookup workloads measures -19% to -33% instructions for the long-route scenarios and parity elsewhere, and `perf record` shows the IC builtins (`KeyedLoadIC`, `LoadIC`) gone from the hot profile.

## Testing

- `npm test` passes: standard, 521 tests via borp, tsd.
- Benchmarked with the repo's own suite as above; per-scenario instruction counts (`perf stat`), cycle profiles (`perf record`), and V8 `--trace-turbo-inlining` used to separate real effects from timing noise and to verify inlining behavior.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01LKNnSRg1hTJsyUZaJYx3QF

### DIFF
--- a/index.js
+++ b/index.js
@@ -103,6 +103,7 @@ function Router (opts) {
 
   this.routes = []
   this.trees = Object.create(null)
+  this._treeGET = null
 }
 
 Router.prototype.on = function on (method, path, opts, handler, store) {
@@ -177,6 +178,9 @@ Router.prototype._on = function _on (method, path, opts, handler, store) {
   }
 
   let currentNode = this.trees[method]
+  if (method === 'GET') {
+    this._treeGET = currentNode
+  }
   let parentNodePathIndex = currentNode.prefix.length
 
   const params = []
@@ -476,6 +480,7 @@ Router.prototype.addConstraintStrategy = function (constraints) {
 
 Router.prototype.reset = function reset () {
   this.trees = Object.create(null)
+  this._treeGET = null
   this.routes = []
 }
 
@@ -573,8 +578,12 @@ Router.prototype.callHandler = function callHandler (handle, req, res, ctx) {
 }
 
 Router.prototype.find = function find (method, path, derivedConstraints) {
-  let currentNode = this.trees[method]
-  if (currentNode === undefined) return null
+  // GET is by far the most common method. Comparing two interned strings is
+  // a pointer comparison, and _treeGET is a plain instance field whose load
+  // compiles to a fixed-offset access, while this.trees is a 35-key
+  // dictionary-mode object whose lookup always goes through a generic IC.
+  let currentNode = method === 'GET' ? this._treeGET : this.trees[method]
+  if (currentNode == null) return null
 
   if (path.charCodeAt(0) !== 47) { // 47 is '/'
     path = path.replace(FULL_PATH_REGEXP, '/')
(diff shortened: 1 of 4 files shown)


ivan-tymoshenko (approved):
lgtm

delvedor (approved):
Awesome!!

PR #449, opened 2026-08-14
fix(router): keep static segments between params in route patterns

rajanpanth:
## Summary

Preserves interior static path segments that appear between dynamic parameters when compiling route patterns.

## Why

Dropping interior static segments can route requests incorrectly for mixed static/param patterns.


rajanpanth:
Closing this duplicate: this exact commit was already merged in #442. Sorry for the duplicate submission.

PR #451, opened 2026-08-21
fix: parse absolute-form request targets safely

mcollina:
## Summary

- replace regex-based absolute-form path extraction with component-aware HTTP(S) URL handling
- normalize an empty absolute-form path to `/` without treating query slashes as path data
- preserve the original query string and route invalid absolute-form targets through `onBadUrl`
- cover direct lookup, router dispatch, and a real Node.js HTTP server request

## Validation

- `npm test`
- `git diff --check`


delvedor (approved):
Less regexes for a better world!

